PR #67, opened 2020-09-27
Async refactor and Serverless support

benweissmann:
- Rework the async flow to use async/await instead of explicit promises
- Break the express app out into a separate file from the http server
- Add a AWS lambda entrypoint and a serverless.yml config

I've done some testing and it appears to be working -- but I definitely didn't test all the nooks and crannies! You should have full read-write access to this fork for feel free to just push changes/fixes directly to this branch -- I'll be mostly offline tomorrow.


tomerovadia:
Wonderful, thank you so so much for doing this! I skimmed over it and it's looking good. I will take a close look tomorrow before Yom Kippur, but may need a bit more time in which case I'll continue Monday evening. I might also take Tuesday off of work so I can work on this and the other tasks more.

benweissmann:
Yup, no rush -- gmar tov!

router.js
```diff
@@ -209,154 +223,169 @@ const routeVoterToSlackChannelHelper = (userInfo, redisClient, twilioPhoneNumber
...
   });
+
+  if (logDebug) console.log(`ROUTER.routeVoterToSlackChannel: Returning voter back to *${destinationSlackChannelName}* from *${adminCommandParams.previousSlackChannelName}*. Voter has been here before.`);
+
+  await SlackApiUtil.sendMessage(
+    `*Operator:* Voter *${userId}* was routed from *${adminCommandParams.previousSlackChannelName}* back to this channel by *${adminCommandParams.routingSlackUserName}*. See their thread with *${twilioPhoneNumber}* above.`,
+    {channel: destinationSlackChannelId}
+  );
+
+  const timestampOfLastMessageInThread = DbApiUtil.getTimestampOfLastMessageInThread(userInfo[destinationSlackChannelId]);
```
tomerovadia, line 372: Should there be an `await` here?

PR #71, opened 2020-09-30
Add Postgres SQL scripts for type & table creation, as well as adding indexes and changing varchar columns to text

tomerovadia:


sql/20200927.sql
```diff
@@ -0,0 +1,63 @@
...
+CREATE TYPE voter_status AS ENUM ('UNKNOWN', 'UNREGISTERED', 'REGISTERED', 'REQUESTED_BALLOT', 'RECEIVED_BALLOT', 'IN_PERSON', 'VOTED', 'REFUSED', 'SPAM');
+
+CREATE TABLE messages (
+    message text,
+    entry_point entry_point,
+    direction message_direction,
+    is_demo boolean,
+    automated boolean,
+    successfully_sent boolean,
+    from_phone_number varchar(12),
```
benweissmann, line 12: One recommendation: in Postgres, it's no more efficient to use `varchar(x)` than `text` -- in fact, it's slightly faster to just use `text` because `varchar(x)` is just a `text` field with a constraint to limit the length (postgres uses the same underlying storage mechanism). More details: https://stackoverflow.com/a/4849030/2005979 -- but tl;dr, with Postgres its generally best to just use `text` for text fields regardless of length.

benweissmann (approved):
Looks good -- just one suggestion about postgres text types. Definitely not a blocker, but you might want to add a new file that `ALTER TABLE`s those `varchar` fields to `text` fields for more flexibility.

sql/20200927.sql
tomerovadia: Sounds great to me. I added a 20200930 file with these changes, please take a look.

benweissmann (approved):
Looks great!

PR #73, opened 2020-10-01
Add automated STOP logic

tomerovadia:
In addition to the existing ability for volunteers to manually set a voter to REFUSE so that no texts are sent to them, this PR adds the following automated logic:

1. Initial messages that match STOP are not replied to with an automated welcome or received by Slack
2. Subsequent messages from a voter who initiated conversation with STOP are not replied to or received by Slack. This cannot be undone other than by manually accessing Redis. With no thread in Slack, there is no way for volunteers to send messages to these voters.
3. Any non-initial STOP messages sent from a voter will automatically update the voter's status to REFUSED, both in Postgres and in Slack, and subsequent automated or volunteer messages to them will be blocked unless this action is undone. This STOP message is relayed to Slack so that the volunteer knows what is going on, and subsequent messages are relayed to Slack as well so the volunteer has a full view of that voter's interactions and experience.
4. Adding to the existing logic associated with the manual REFUSED button available to volunteers, pressing this button disables not just volunteer messages to the voter but also automated messages (in case a voter repeatedly texts while in the lobby that they don't want a response).

### DIFF
--- a/sql/20200930-2.sql
+++ b/sql/20200930-2.sql
@@ -1,0 +1,27 @@
+-- Remove "originating" from some column names.
+-- These columns identify the voter's Slack channel and thread,
+-- which in the case of a voter status button press, is necessarily
+-- the same as the originating Slack channel and thread.
+-- Now that the change of a voter status can be automated, these
+-- columns are useful to help identify the voter, even if the "origin"
+-- of the update isn't the channel/thread themselves.
+-- Leaving " originating" for originating_slack_user_name and
+-- originating_slack_user_id.
+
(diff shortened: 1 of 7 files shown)


benweissmann (approved):
Looks good! Just looks like there are a couple broken tests because of the column renames.

tomerovadia:
> Looks good! Just looks like there are a couple broken tests because of the column renames.

Thanks! Fixed. Looks like many were broken due to the changed input requirements to Router functions and TwilioApiUtil.sendMessage.

PR #74, opened 2020-10-02
Create custom voter flow, automated messages & U.S. state options for VoteAmerica

### LINKED ISSUE #58: Create flow with no disclaimer and for any U.S. state selection

### LINKED ISSUE #59: Eliminate need for Slack channel name to ID lookup

### LINKED ISSUE #64: Create support for a national catch-all channel

### LINKED ISSUE #65: Limit the number of U.S. state clarifications
Record in Redis how many attempts were made to understand a user's U.S. state, and consider the voter cleared if either a U.S. state were determined OR if the number of attempts hit the limit. If the number of attempts is exactly the limit, send a last automated message to the voter saying a volunteer is being sought

tomerovadia:
Also:
-add functionality to limit the number of U.S. state clarification automated messages, routing to a National channel if state is not determined
-add functionality to auto-update Redis cache of Slack channel names to channel IDs lookup if a channel is not found when routing a voter (either automatic or admin command)
-remove any processing of user messages that are or follow a STOP message, with the exception of passing to Slack and writing to Postgres


tomerovadia:
Fuzzy- this is ready, except that I can modify it to accommodate your Slack handlers PR changes so that the conflict is easier to manage. What do you think?

benweissmann (approved):
This looks great! Happy to rebase my stuff on top of this if you merge first, or I can merge and you can rebase. Whatever's easiest.

src/slack_api_util.ts
```diff
@@ -237,3 +243,49 @@ export async function fetchSlackMessageBlocks(
     return null;
   }
 }
+
+export async function fetchSlackChannelNamesAndIds(): Promise<SlackChannelNamesAndIds | null> {
+  logger.info(`ENTERING SLACKAPIUTIL.fetchSlackChannelNamesAndIds`);
+  const response = await axios.get('https://slack.com/api/conversations.list', {
```
benweissmann, line 249: This is a paginated endpoint -- I think you need to fetch additional pages if the response has a `response_metadata.next_cursor` property

benweissmann (changes requested):
Sorry, missed this before -- just one small change to the new slack interaction

src/slack_api_util.ts
tomerovadia: (Discussed offline: This is a necessary change, and we'll follow-up with a fix in a separate PR).

PR #76, opened 2020-10-04
`authed_users` check isn't necessary

fongandrew:
Small fix here. This `authed_users` check doesn't do what the comment suggests it does. `authed_users` is a list of (probably human) users who have installed the app, not a list of the actual user IDs mentioned. So for me at least, this condition was always returning false.

In any event, the `slackAppMentionEventHandler` also checks the Slack bot is the first user mentioned, so this doesn't do anything other than possibly saving a few background tasks from being enqueued.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -537,7 +537,12 @@ app.post(
         retryCount,
         retryReason,
       });
-    } else if (reqBody.event.type === 'app_mention' && !retryReason) {
+    } else if (
+      reqBody.event.type === 'app_mention' &&
+      // Require that the Slack bot be the (first) user mentioned.
+      reqBody.event.text.startsWith(`<@${process.env.SLACK_BOT_USER_ID}>`) &&
+      !retryReason
+    ) {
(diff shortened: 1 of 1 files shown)


tomerovadia (changes requested):
Great catch! This actually might fix this command, which Fuzzy noticed is broken in the production workspace because of the deprecation of authed_users: https://api.slack.com/changelog/2020-09-15-events-api-truncate-authed-users

Question: you mentioned that `slackAppMentionEventHandler` already does this check for a bot mention -- can you point to where that is?

fongandrew:
`slackAppMentionEventHandler` calls `Router.handleSlackAdminCommand` which calls `CommandUtil.parseSlackCommand` which checks here: https://github.com/tomerovadia/VoterHelpLine/blob/2276bc0794b2a3847f9548e8a4d2e5a27fce432d/src/command_util.ts#L141-L152

Although, now that I think about it, that means you'll get an error message if the mention isn't the first word, rather than a noop. Lemme fix that.

tomerovadia (approved):
Thanks for catching/fixing this!

PR #77, opened 2020-10-04
Add Twilio message callbacks

### LINKED ISSUE #63: Slack messages intended to be routed to a voter are tagged with an emoji indicating whether or not they were successfully routed

benweissmann:
This PR does a couple things:

- Specify a callback URL when sending a message to Twilio, so we get notified about delivery success/failure of our text messages
- Handle these callbacks by updating the DB and adding a reaction to the Slack message
- Store message timestamps from Slack as strings, not numbers. This is important because right now a message TS of `1234.567000` from slack gets stored as `1234.567`. But if you try to pass `1234.567` to the slack API when you're trying to do something with that message, Slack will say that's an invalid timestamp -- the timestamps are supposed to be strings with the same number of digits of precision as slack tell you, so dropping the trailing 0's makes it invalid. See this for details: https://github.com/slackhq/slack-api-docs/issues/7#issuecomment-67913241

**IMPORTANT**: this now requires the bot to have the `reactions:write` permission. I've added this to the VoteAmerica staging and prod instances, but you'll have to add it to your dev instances to test this out.

### DIFF
--- a/sql/20201004.sql
+++ b/sql/20201004.sql
@@ -26,6 +26,18 @@ ADD COLUMN twilio_callback_timestamp timestamptz;
 ALTER TABLE messages
 ALTER COLUMN slack_parent_message_ts TYPE text USING trim(to_char(slack_parent_message_ts, '99999999999999999999999999.999999'));
 
-
 ALTER TABLE messages
 ALTER COLUMN slack_message_ts TYPE text USING trim(to_char(slack_message_ts, '99999999999999999999999999.999999'));
+
+ALTER TABLE voter_status_updates
+ALTER COLUMN slack_parent_message_ts TYPE text USING trim(to_char(slack_parent_message_ts, '99999999999999999999999999.999999'));
+
+ALTER TABLE voter_status_updates
+ALTER COLUMN action_ts TYPE text USING trim(to_char(action_ts, '99999999999999999999999999.999999'));
+
+ALTER TABLE volunteer_voter_claims
+ALTER COLUMN slack_parent_message_ts TYPE text USING trim(to_char(slack_parent_message_ts, '99999999999999999999999999.999999'));
+
+ALTER TABLE volunteer_voter_claims
+ALTER COLUMN action_ts TYPE text USING trim(to_char(action_ts, '99999999999999999999999999.999999'));
+
--- a/sql/20201004.sql
+++ b/sql/20201004.sql
@@ -1,0 +1,31 @@
+-- couple more indexes for message
+
+CREATE INDEX idx_messages_slack_parent_message_ts
+ON messages(slack_parent_message_ts);
+
+CREATE INDEX idx_messages_twilio_message_sid
+ON messages(twilio_message_sid);
+
+-- add fields to messages to store Twilio callback data
+
+ALTER TABLE messages
+ADD COLUMN twilio_callback_status text;
+
+
+ALTER TABLE messages
+ADD COLUMN twilio_callback_timestamp timestamptz;
+
+-- store slack message timestamps as text, not numbers (they're not really numbers)
+-- See: https://github.com/slackhq/slack-api-docs/issues/7#issuecomment-67913241
+--
+-- When converting, we need to be careful to format the existing timestamps
+-- with 6 decimal places. This is because in the past we were storing a ts
+-- provided by slack as a number, so "1234.567000" was being stored as 1234.567.
+-- In order to now convert it back to the original timestamp, we need to use
+-- 6 decimal places to that it ends up as "1234.567000" and not "1234.567".
+ALTER TABLE messages
+ALTER COLUMN slack_parent_message_ts TYPE text USING trim(to_char(slack_parent_message_ts, '99999999999999999999999999.999999'));
+
+
+ALTER TABLE messages
+ALTER COLUMN slack_message_ts TYPE text USING trim(to_char(slack_message_ts, '99999999999999999999999999.999999'));
--- a/src/db_api_util.ts
+++ b/src/db_api_util.ts
@@ -49,7 +49,7 @@ export type DatabaseVoterStatusEntry = {
   slackChannelName: string | null;
   slackChannelId: string | null;
   slackParentMessageTs: string | null;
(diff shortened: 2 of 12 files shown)


fongandrew (approved):
Pulled and tested on my end. I'll let @tomerovadia take a look, but it looks good to me.

sql/20201004.sql
```diff
@@ -0,0 +1,31 @@
...
+
+-- store slack message timestamps as text, not numbers (they're not really numbers)
+-- See: https://github.com/slackhq/slack-api-docs/issues/7#issuecomment-67913241
+--
+-- When converting, we need to be careful to format the existing timestamps
+-- with 6 decimal places. This is because in the past we were storing a ts
+-- provided by slack as a number, so "1234.567000" was being stored as 1234.567.
+-- In order to now convert it back to the original timestamp, we need to use
+-- 6 decimal places to that it ends up as "1234.567000" and not "1234.567".
+ALTER TABLE messages
```
tomerovadia, line 26: can you do this for the voter_status_updates and volunteer_voter_claims tables as well?

src/slack_api_util.ts
```diff
@@ -289,3 +289,29 @@ export async function updateSlackChannelNamesAndIdsInRedis(
...
+    {
+      headers: {
+        Authorization: `Bearer ${process.env.SLACK_BOT_ACCESS_TOKEN}`,
+      },
+    }
+  );
+
+  if (!response.data.ok) {
+    throw new Error(
+      `SLACKAPIUTIL.sendMessage: ERROR in sending Slack message: ${response.data.error}`
```
tomerovadia, line 314: Update for this function

tomerovadia: Update for this function

sql/20201004.sql
```diff
@@ -0,0 +1,31 @@
...
+ADD COLUMN twilio_callback_timestamp timestamptz;
+
+-- store slack message timestamps as text, not numbers (they're not really numbers)
+-- See: https://github.com/slackhq/slack-api-docs/issues/7#issuecomment-67913241
+--
+-- When converting, we need to be careful to format the existing timestamps
+-- with 6 decimal places. This is because in the past we were storing a ts
+-- provided by slack as a number, so "1234.567000" was being stored as 1234.567.
+-- In order to now convert it back to the original timestamp, we need to use
+-- 6 decimal places to that it ends up as "1234.567000" and not "1234.567".
```
tomerovadia, line 25: Can you do this for voter_status_updates and volunteer_voter_claims as well?

src/db_api_util.ts
```diff
@@ -501,3 +501,52 @@ export async function logVolunteerVoterClaimToDb(
...
+      console.log({
+        slackChannel: row.slack_channel,
+        slackMessageTs: row.slack_message_ts,
+      });
+      return {
+        slackChannel: row.slack_channel,
+        slackMessageTs: row.slack_message_ts,
+      };
+    } else {
+      logger.error(`DBAPIUTIL.getLatestVoterStatus: No voter status for user`);
```
tomerovadia: Update

src/db_api_util.ts
```diff
@@ -501,3 +501,52 @@ export async function logVolunteerVoterClaimToDb(
...
+      RETURNING
+        slack_channel,
+        slack_message_ts
+    `,
+      [status, messageSid]
+    );
+
+    if (result.rows.length > 0) {
+      const row = result.rows[0];
+      console.log({
```
tomerovadia: Should this use logger?

tomerovadia (changes requested):
All looks excellent. Added a few comments.

src/twilio_status_callback_handler.ts
```diff
@@ -0,0 +1,51 @@
...
+  // Twilio delivers callbacks for lots of intermediate states (queued, sent
+  // to carrier, etc.). We just care about the final status of the message,
+  // so we don't update Postgres for those intermediate statuses.
+  if (!(messageStatus in FINAL_STATUSES)) {
+    logger.info(`Twilio status is non-final; not processing`);
+    return;
+  }
+
+  // Update the Postgres DB with the message status and current timestamp
+  const slackMessageInfo = await logTwilioStatusToDb(messageSid, messageStatus);
```
tomerovadia, line 30: Out of curiosity, why doesn't this need a catch? I see the function itself has a try/finally within it like other DBApiUtil functions, but I think we have try/catches around calls to other DBApiUtil functions elsewhere.

src/db_api_util.ts
```diff
@@ -501,3 +501,52 @@ export async function logVolunteerVoterClaimToDb(
...
+  logger.info(`ENTERING DBAPIUTIL.logTwilioStatusToDb`);
+
+  const client = await pool.connect();
+
+  try {
+    const result = await client.query(
+      `
+      UPDATE messages
+      SET
+        twilio_callback_status = $1,
```
tomerovadia, line 524: What are your thoughts on modifying the successfully_sent field as well?

src/twilio_status_callback_handler.ts
benweissmann: We only need a try/catch if there's something we want to do when there's an error other than just return a 500. In this case, there's nothing to do -- if we can't query the database, we can't get the slack message info, so we can't add a reaction. So we might as well just let the error bubble up and return a 500 -- try/catch is just for when there's some useful fallback behavior or graceful failure mode.

src/db_api_util.ts
benweissmann: I think we can leave `successfully_sent` as is -- to indicate that we sent the message off to Twilio. I'm not sure there's much use in updating both successfully_sent *and* the twilio_status when we get a twilio status callback.

tomerovadia (approved):

src/twilio_status_callback_handler.ts
tomerovadia: What about catching so we can log to Sentry?

src/db_api_util.ts
tomerovadia: Ok great point

PR #79, opened 2020-10-05
Create reset demo shortcut

### LINKED ISSUE #30: Log all admin actions

### LINKED ISSUE #60: Create automated way to clear volunteer practice Redis data

### LINKED ISSUE #61: Create a way to archive demo messages in Postgres so that they are not included in message history when a practicing volunteer is routed between channels
Nightly, all demo messages can be archived so that they are no longer used in message history when routing a voter between channels.

tomerovadia:
Primary changes:
1. Create a Reset Voter Slack Shortcut that allows volunteers to reset a voter from that voter's thread's parent message. Doing so deletes the voter's lookups in Redis (the Twilio-to-Slack userInfo as well as all Slack-to-Twilio redisData lookups) and archives all the voter's messages with this phone number in Postgres (so they are not included in subsequent demo message histories).
2. Create `commands` table in Postgres to keep track of commands issued from Slack (admin or not). Starting only with this RESET_DEMO command.

Additional changes:
1. Add an unused redis_initialization.ts file for developers to seed Redis with open pod data.
2. Create an `external_configs/` directory with a `slack.md` to include Slack configurations and keep track of changes that need to be made to different environments. If you make a change in your Slack config that needs to be pushed to other environments, please include it.
3. Fix a bug in SlackApiUtil.fetchSlackChannelNamesAndIds (which was mapping from ids to names instead of vice versa).

**IMPORTANT**: Please add "commands" to bot scope in your local environments and also add the Reset Voter Slack Shortcut. See `external_configs/slack` for additional information..


src/async_jobs.ts
```diff
@@ -31,46 +36,97 @@ async function slackInteractivityHandler(payload: SlackEventPayload) {
...
+            twilioPhoneNumber: redisData ? redisData.twilioPhoneNumber : null,
+          });
+        }
+        return;
+      }
+      case 'message_action': {
+        logger.info(
+          `SERVER POST /slack-interactivity: Determined user interaction is a message shortcut.`
+        );
+        if (payload.callback_id === 'reset_demo') {
```
fongandrew, line 95: Nit: In https://github.com/tomerovadia/VoterHelpLine/pull/84, I created a new module + enum for these `callback_id`. It'll be nice to consolidate these in one place.

src/slack_interaction_handler.ts
```diff
@@ -319,3 +344,201 @@ export async function handleVolunteerUpdate({
...
+  console.log(modalPrivateMetadata);
+  const redisUserInfoKey = `${modalPrivateMetadata.userId}:${modalPrivateMetadata.twilioPhoneNumber}`;
+
+  const slackThreads = (await DbApiUtil.getSlackThreadsForVoter(
+    modalPrivateMetadata.userId,
+    modalPrivateMetadata.twilioPhoneNumber
+  )) as SlackThreadInfo[];
+
+  console.log('slackThreads');
+  console.log(slackThreads);
```
fongandrew: Should these console logs be here? There are a handful in this function.

src/slack_interaction_handler.ts
```diff
@@ -319,3 +344,201 @@ export async function handleVolunteerUpdate({
     newBlocks: payload.message.blocks,
   });
 }
+
+// This function receives the initial request to reset a demo
+// and response by creating a modal populated with data needed
+// to reset the demo if the Slack user confirms.
```
fongandrew, line 350: Maybe I'm missing it, but is there (should there be?) an admin check in this function? Otherwise it seems like anyone can reset demos?

src/slack_interaction_handler.ts
```diff
@@ -319,3 +344,201 @@ export async function handleVolunteerUpdate({
...
+
+    // Store the relevant information in the modal so that when the requested action is confirmed
+    // the data needed for the necessary actions is available.
+    slackView = SlackBlockUtil.resetConfirmationSlackView(
+      commandType,
+      modalPrivateMetadata
+    );
+  }
+
+  await SlackApiUtil.renderModal(payload.trigger_id, slackView);
```
fongandrew, line 437: One potential issue here: The `trigger_id` expires after 3 seconds, and we're doing this work inside a background task after a DB command. You'll get here within 3 seconds _most_ of the time, but if there's a large amount of the traffic / network congestion / etc., the command might fail.

This isn't a part of Slack's API that I particularly love, but the recommended way to handle this is to open the modal ASAP in a loading state and then to update the modal with stuff as needed. This is fine for now (and it looks like it's currently processing under 3s so maybe it's fine?), but I'm working on some utils on my end that'll make the open-then-update flow a little nicer.

src/db_api_util.ts
```diff
@@ -69,13 +72,30 @@ export type DatabaseVolunteerVoterClaim = {
...
+      originatingSlackUserName: string | null;
+      originatingSlackUserId: string | null;
+      slackChannelName: string | null;
+      slackChannelId: string | null;
+      slackParentMessageTs: string | null;
+      success: boolean | null;
+      actionTs: string | null;
+      failureReason?: string | null;
+    }
+  | SlackModalPrivateMetadata;
```
benweissmann: nit: I don't think you need this line -- `SlackModalPrivateMetadata` is compatible with the type above, so you should be able to just pass a SlackModalPrivateMetadata as a DatabaseCommandEntry (Typescript has a structural type system -- so it doesn't care what your types are named or if two objects have nominally different types, as long as those types are structurally the same). You might need to make `success` optional here (`success?: boolean | null`) to make it fully compatible with SlackModalPrivateMetadata.

src/slack_interaction_handler.ts
tomerovadia: Removed. Thanks for catching!

src/slack_interaction_handler.ts
tomerovadia: Thanks for helping me think about this! For this particular command, we want all volunteers to be able to use it to reset their own demo conversations. With regard to any _particular_ demo conversation, right now only those Slack users who have access to the conversation can reset it, which should only be admins, the volunteer in the demo conversation and their podmates. I think that's perfect in this case. And if someone selects Reset Demo from a thread that is no longer active or tries to Reset Demo on a non-demo conversation, they will get an error modal. What do you think / does that sound right to you? Really appreciate you helping me think about this as I've had my head down in the weeds on this for a while.

With regard to permissions for the other commands we have lined up:
-**Route Voter To Channel:** should be a Shortcut that only works for admins
-**Opening Channels (however we implement it):** only available to admins
-**Route Voter to Journey:** should have the same access control as Reset Voter

src/slack_interaction_handler.ts
tomerovadia: Great point, I was worried about that and hadn't thought of that solution. Since we're looking to start volunteer practice today and it does appear to be working for now (and if it failed, luckily it would only affect volunteers and not voters), what are your thoughts on submitting this change and following up with a fix?

src/async_jobs.ts
tomerovadia: That's awesome, thanks for doing that! I'm about to hop over to look at all the changes you sent.

fongandrew (approved):
Looks good!

benweissmann (commented):
re: andrew's comment on the trigger_id -- I think we are likely to hit that case pretty frequently for our staging/production environment on Lambda. I think we should try and get that resolved in the next day or two.

src/db_api_util.ts
tomerovadia: Done thanks!

tomerovadia:
> re: andrew's comment on the trigger_id -- I think we are likely to hit that case pretty frequently for our staging/production environment on Lambda. I think we should try and get that resolved in the next day or two.

Sounds good to me.

benweissmann (approved):

PR #84, opened 2020-10-05
Use action_id for Slack actions

fongandrew:
Anything we do in a Slack modal will come through the same interactivity endpoint used for the voter status updates on messages. Currently it's doing some stuff based on the action type and value, but those might not always be unique. The recommended way to distinguish these is to use some combination of a unique `block_id` or `action_id`.

This just defines some static `action_id`s for the voter status stuff and applies it to the blocks created for each message. It also updates `populateDropdownNewInitialValue` to use `action_id`s when swapping out initial values in replacement blocks. I'll work on creating a separate code path to handle incoming modal actions based on those action IDs in a separate PR.

Also, does some stuff around tests and fixtures. More comments inline. Recommend reviewing by commit.


.eslintrc
```diff
@@ -27,8 +27,9 @@
   },
   "overrides": [
     {
-      "files": [ "**/*.js" ],
+      "files": ["**/*.js"],
       "rules": {
+        "@typescript-eslint/explicit-module-boundary-types": "off",
```
fongandrew, line 32: I was writing some of the test fixtures using plain JS to avoid having to type them and ran into this rule. Happy to switch the test fixtures to TS, but I noticed the existing tests were already in JS, so it seems like that's how we're rolling here.

jest.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  // Restores all mock / spied-on implementations before each test
+  restoreMocks: true,
```
fongandrew, line 3: This restores any `jest.spyOn` calls after each test. Open to whatever you'd prefer here, but my experience in the past with Jest has been:
* Re-creating the mock before each test isn't a big deal most of the time
* Forgetting to reset the mock after a test can lead to confusing test failures

src/fixtures/slack_action_payload.js
```diff
@@ -0,0 +1,31 @@
+export const makeSlackMessageBlockActionPayload = (props = {}) => ({
```
fongandrew, line 1: Note that unlike the blocks defined in the block utils, we generate these with factory functions because any mutations will mess with test isolation.

src/slack_interaction_handler.ts
```diff
@@ -197,10 +198,19 @@ export async function handleVoterStatusUpdate({
...
-      SlackBlockUtil.populateDropdownNewInitialValue(
-        payload.message.blocks,
-        selectedVoterStatus as VoterStatus
-      );
+      if (
+        !SlackBlockUtil.populateDropdownNewInitialValue(
+          payload.message.blocks,
+          payload.actions
+            ? payload.actions[0].action_id
+            : SlackActionId.VOTER_STATUS_DROPDOWN,
```
fongandrew, line 231: I ran into issues when testing because messages created before we started using static action IDs stopped working. Those old blocks do have Slack-auto-generated `action_id`s and they should be specified in the payload though (the type says `payload.actions` is sometimes undefined, hence this ternary), so we can try to use those `action_id`s in the payload if available.

src/fixtures/slack_voter_status_message.js
```diff
@@ -0,0 +1,218 @@
...
+  latest_reply: '1601873059.004600',
+  reply_users: ['U01BGEC9T7H', 'U01BGDZHPDM'],
+  subscribed: true,
+  last_read: '1601873059.004600',
+});
+
+// Voter status message object with action IDs
+export const makeVoterStatusMessageWithActionIds = () => {
+  const message = makeVoterStatusMessageWithoutActionIds();
+  message.blocks[1].elements[0].action_id = 'VOLUNTEER_UPDATE';
```
tomerovadia: nit: Since the volunteer update is also a dropdown, should we name it VOLUNTEER_DROPDOWN to match VOTER_STATUS_DROPDOWN? And is it helpful to have the others also describe the block element type (e.g. VOTER_STATUS_VOTED_BUTTON, VOTER_STATUS_REFUSED_BUTTON, VOTER_STATUS_SPAM_BUTTON)?

I don't feel passionately about this so feel free to ignore.

src/slack_interaction_ids.ts
```diff
@@ -0,0 +1,11 @@
+// List of enums
+
+// Action IDs are used in interactive blocks
+export enum SlackActionId {
+  CLOSED_VOTER_PANEL_UNDO = 'CLOSED_VOTER_PANEL_UNDO',
```
tomerovadia: Is this also defined in slack_voter_status_message? Can we have that file import these values so they're defined in one place?

src/slack_block_util.ts
```diff
@@ -306,34 +322,34 @@ export function replaceVoterPanelBlocks(
...
-        if (isVoterStatusOption) {
+        if (element.action_id === actionId) {
           if (element.type === 'static_select') {
-            element.initial_option.value = newInitialValue;
-            element.initial_option.text.text =
-              voterStatusOptions[newInitialValue];
-            // Javascript modifies the blocks by reference, so end but don't return anything.
-            return;
+            // Assume new options is already in the list of old options
+            element.initial_option = element.options.find(
```
tomerovadia: If `newInitialValue` is null, will the `initial_option` end up being set to null, and this function would return true?

src/slack_interaction_handler.ts
tomerovadia: Sounds good, this makes sense. If `payload.actions` is sometimes undefined, do you know if it's at least consistent?

.eslintrc
tomerovadia: @benweissmann is the expert here

jest.config.js
tomerovadia: Sounds good to me. I think I hacked together something like this in router.test.js.

src/fixtures/slack_voter_status_message.js
```diff
@@ -0,0 +1,218 @@
...
+  reply_count: 21,
+  reply_users_count: 2,
+  latest_reply: '1601873059.004600',
+  reply_users: ['U01BGEC9T7H', 'U01BGDZHPDM'],
+  subscribed: true,
+  last_read: '1601873059.004600',
+});
+
+// Voter status message object with action IDs
+export const makeVoterStatusMessageWithActionIds = () => {
```
tomerovadia, line 212: This is only used in the test, correct?

fongandrew: Yup, it's only in the test

src/slack_interaction_handler.ts
fongandrew: Yup, it's consistent.

src/slack_block_util.ts
tomerovadia: If we don't expect newInitialValue to ever be null, should we have a fallback to leave the `initial_option` as-is and maybe throw an error that logs to Sentry (and have the function return false)?

tomerovadia (approved):

tomerovadia:
@fongandrew do you think this PR might fix this issue?: https://github.com/tomerovadia/VoterHelpLine/issues/90

fongandrew:
Unclear -- I'm not sure they're related.

tomerovadia (approved):

.eslintrc
benweissmann: Yup, having the testing code in JS seems fine to me

benweissmann (approved):
Looks great!

PR #88, opened 2020-10-05
Open Slack modal immediately and update asynchronously

benweissmann:
@fongandrew Is this roughly what you had suggested to handle the 3-second timeout?

@tomerovadia -- I made a few updates to the `slack_api_utils`:
- Instead of checking for NODE_ENV == test, I updated `package.json` to set the logging level to FATAL in tests so we don't see any log messages (you can change this back to `info` or `debug` when debugging tests)
- I created a custom Axios instance that handles the Slack base URL and the headers for POST requests. `Content-Type` doesn't need to be in the request body (it's a header, not part of the request body) so I removed that from all of the API calls.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -607,7 +607,10 @@ app.post(
 
     const metadata: InteractivityHandlerMetadata = {};
 
-    if (payload.type === 'message_action') {
+    if (
+      payload.type === 'message_action' &&
+      payload.callback_id === 'reset_demo'
+    ) {
       // For message actions, we always show a confirmation modal. Because we
       // have to show this modal within 3 seconds, we immediately make the call
       // to show a loading state, and then pass the modal ID on to the async
--- a/src/app.ts
+++ b/src/app.ts
@@ -15,11 +15,16 @@ import * as LoadBalancer from './load_balancer';
 import * as SlackUtil from './slack_util';
 import * as TwilioUtil from './twilio_util';
 import * as SlackInteractionApiUtil from './slack_interaction_api_util';
+import * as SlackBlockUtil from './slack_block_util';
 import logger from './logger';
 import redisClient from './redis_client';
 import { EntryPoint, Request, UserInfo } from './types';
-import { enqueueBackgroundTask } from './async_jobs';
+import {
+  enqueueBackgroundTask,
+  InteractivityHandlerMetadata,
+} from './async_jobs';
 import { handleTwilioStatusCallback } from './twilio_status_callback_handler';
+import { SlackInteractionEventPayload } from './slack_interaction_handler';
 
 const app = express();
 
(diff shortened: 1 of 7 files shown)


src/slack_api_util.ts
```diff
@@ -316,38 +292,58 @@ export async function addSlackMessageReaction(
...
-      trigger_id: triggerId,
-      view,
-    },
-    {
-      headers: {
-        Authorization: `Bearer ${process.env.SLACK_BOT_ACCESS_TOKEN}`,
-      },
-    }
-  );
+  const response = await slackAPI.post('https://slack.com/api/views.open', {
```
tomerovadia: Should this be simply `views.open`?

src/slack_api_util.ts
```diff
@@ -316,38 +292,58 @@ export async function addSlackMessageReaction(
...
+
+/**
+ * Updates a modal in slack given the modal ID from renderModal
+ */
+export async function updateModal(
+  viewId: string,
+  view: SlackView
+): Promise<string> {
+  logger.info(`ENTERING SLACKAPIUTIL.updateModal`);
+  const response = await slackAPI.post('https://slack.com/api/views.update', {
```
tomerovadia: Should this be simply `views.update`?

src/slack_block_util.ts
```diff
@@ -18,15 +18,13 @@ export type SlackView = {
...
       type: string;
-      text: {
-        type: string;
-        text: string;
-      };
-    }
-  ];
+      text: string;
+    };
+  }[];
```
tomerovadia, line 27: Cool, good to know this is a thing!

src/slack_block_util.ts
```diff
@@ -249,6 +247,17 @@ export const voterStatusPanel: SlackBlock = {
   ],
 };
 
+export function loadingSlackView(): SlackView {
+  return {
+    title: {
+      type: 'plain_text',
+      text: 'Loading...',
+    },
+    blocks: [],
```
fongandrew: We should probably have a single section block with some loading text in it. I think Slack does funny things when there are no blocks at all.

src/app.ts
```diff
@@ -598,9 +603,22 @@ app.post(
       return;
     }
 
-    const payload = JSON.parse(req.body.payload);
+    const payload: SlackInteractionEventPayload = JSON.parse(req.body.payload);
 
-    await enqueueBackgroundTask('slackInteractivityHandler', payload);
+    const metadata: InteractivityHandlerMetadata = {};
+
+    if (payload.type === 'message_action') {
```
fongandrew: Can we add check the `callback_id` instead of just looking at `payload.type`? We may be adding additional message actions in the future that don't open a modal.

src/slack_interaction_handler.ts
```diff
@@ -434,7 +436,7 @@ export async function receiveResetDemo({
     );
   }
 
-  await SlackApiUtil.renderModal(payload.trigger_id, slackView);
+  await SlackApiUtil.updateModal(viewId, slackView);
```
tomerovadia: Previously, if an error was thrown above (Interaction received for voter who has redisData but not userInfo), then nothing happens, which isn't ideal (preferably we'd show a modal). After this change, the Loading... modal would appear but never be updated. Can we add an updateModal call right before throwing an error? And perhaps a "Close" to the Loading... modal just in case?

src/async_jobs.ts
```diff
@@ -92,6 +95,14 @@ async function slackInteractivityHandler(
         logger.info(
           `SERVER POST /slack-interactivity: Determined user interaction is a message shortcut.`
         );
+
+        const { viewId } = interactivityMetadata;
```
tomerovadia, line 99: Why is this necessary if all it's carrying is a viewId (ie why don't you just pass the viewId)? In case we want to carry more than the viewId in the future?

src/slack_api_util.ts
```diff
@@ -8,6 +8,14 @@ import { SlackBlock, SlackView } from './slack_block_util';
 import * as RedisApiUtil from './redis_api_util';
 import { PromisifiedRedisClient } from './redis_client';
 
+const slackAPI = axios.create({
```
tomerovadia, line 11: Your edits to this file are great, thanks for doing this.

tomerovadia (changes requested):

src/slack_interaction_handler.ts
fongandrew: Slack automatically adds a close button if you don't specify one, so a user wouldn't be stuck. But updating the modal to say "whoops, there was an error" would be the user-friendly thing to do.

src/async_jobs.ts
benweissmann: Yup, that was my thinking.

tomerovadia (approved):

fongandrew (approved):

Issue #95, opened 2020-10-07
Brenna entered an endless loop of prompts


tomerovadia:
DataDog shows the error line: `SLACKAPIUTIL.sendMessage: ERROR in sending Slack message: restricted_action`
Sentry shows the error as well: https://sentry.io/organizations/voteamerica/issues/1942637736/events/c9bdba36450e427bae6ab3acb9183e10/

We should Sentry and/or log more of the error here: https://github.com/tomerovadia/VoterHelpLine/blob/3e1f8223ff79bb0eb5c72c1687d044dcb29a2354/src/slack_api_util.ts#L75

tomerovadia:
Resolved. Workspace settings were changed to disallow `@channel` for anyone other than admins and workspace owners, which excludes the bot, which uses it to alert members of a channel to new incoming voters. Thanks @fongandrew for figuring this out.

tomerovadia:
Also, the above PR ensures that if this problem were encountered again, there would not be an endless loop, and an error would be thrown so we'd be notified.

PR #96, opened 2020-10-08
Add better logging in SlackApiUtil, and send user an automated error …

tomerovadia:
…text if they select helpline and a thread isn't successfully created for them in Slack

Intended to help debug issues like: https://github.com/tomerovadia/VoterHelpLine/issues/95

### DIFF
--- a/src/slack_api_util.ts
+++ b/src/slack_api_util.ts
@@ -212,9 +212,7 @@ export async function fetchSlackUserName(
     logger.error(
       `SLACKAPIUTIL.fetchSlackUserName: Failed to reveal Slack user name (${userId}). Error: response.data: ${JSON.stringify(
         response.data
-      )} data.response_metadata: ${JSON.stringify(
-        response.data.response_metadata
-      )}.`
+      )}`
     );
     return null;
   }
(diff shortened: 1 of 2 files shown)


benweissmann (approved):

src/slack_api_util.ts
```diff
@@ -198,7 +210,11 @@ export async function fetchSlackUserName(
     return response.data.user.real_name;
   } else {
     logger.error(
-      `SLACKAPIUTIL.fetchSlackUserName: Failed to reveal Slack user name (${userId}). Error: ${response.data.error}.`
+      `SLACKAPIUTIL.fetchSlackUserName: Failed to reveal Slack user name (${userId}). Error: response.data: ${JSON.stringify(
+        response.data
```
fongandrew, line 214: Is stringifying both the `data` and `response_metadata` necessary? Isn't `response_metadata` already included in the data?

tomerovadia: Good catch thanks. Stupid mistake. Should be fixed now.

fongandrew (approved):

PR #97, opened 2020-10-08
Add redis-backed deduplication to outbound sms

### LINKED ISSUE #92: Repeat message due to http_timeout Slack error and two subsequent retries
image (3)

Lauren was demoing the prod environment and received a text message twice. Postgres shows the message three times, and two are retries:

`SELECT * FROM messages WHERE message LIKE '%probably fine%';`

1	"http_timeout"	"delivered"	"2020-10-06 22:59:30.964548+00"
2	"http_timeout"	"delivered"	"2020-10-06 23:00:30.485861+00"
3      null	                         "delivered"     "2020-10-06 22:59:28.193453+00"

Sentry caught this too, via the emoji reactions: `SLACKAPIUTIL.addSlackMessageReaction: ERROR in adding reaction: already_reacted`
https://sentry.io/organizations/voteamerica/issues/1940321582/events/f4d405441c9b445fa1e1bd7a54aa3b2f/

benweissmann:
- Pass a unique slack message ID to the twilio sendMessage function to use as a deduplication ID
- Deduplicate outbound twilio messages when a deduplication ID is provided

### DIFF
--- a/src/deduplication.ts
+++ b/src/deduplication.ts
@@ -11,7 +11,7 @@ const DEDUPE_WINDOW_SECONDS = 60 * 60;
  * If there's an error talking to redis, it will return true (erring on the
  * side of permitting duplication if there's a problem with Redis).
  */
-export default async function deduplicate(key: string): Promise<boolean> {
+export default async function isFirstUseOfKey(key: string): Promise<boolean> {
   try {
     const res = await deduplicationRedisClient.setAsync(
       `deduplication:${key}`,
@@ -21,11 +21,7 @@ export default async function deduplicate(key: string): Promise<boolean> {
       'NX'
     );
 
-    if (res) {
-      return true;
-    } else {
-      return false;
-    }
+    return !!res;
   } catch (err) {
     logger.warn(`DEDUPLICATION: failed deduplicate message with key ${key}`);
     logger.warn(err);
--- a/src/deduplication.ts
+++ b/src/deduplication.ts
@@ -1,0 +1,35 @@
+import { deduplicationRedisClient } from './redis_client';
+import * as Sentry from '@sentry/node';
+
+const DEDUPE_WINDOW_SECONDS = 60 * 60;
+
+/**
+ * Implements redis-backed deduplication. The first time this is called with
+ * a given key, it will return true. After that, it will return false until
+ * the key expires in 1 hour.
+ *
+ * If there's an error talking to redis, it will return true (erring on the
+ * side of permitting duplication if there's a problem with Redis).
+ */
+export default async function deduplicate(key: string): Promise<boolean> {
+  try {
+    const res = await deduplicationRedisClient.setAsync(
+      `deduplication:${key}`,
+      'true',
+      'EX',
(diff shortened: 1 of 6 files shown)


src/redis_client.ts
```diff
@@ -31,3 +35,15 @@ export type PromisifiedRedisClient = typeof redisClient & {
 };
 
 export default redisClient as PromisifiedRedisClient;
+
+const deduplicationRedisClient = redis.createClient(
+  (process.env.REDISCLOUD_URL_DEDUPLICATION ||
```
tomerovadia, line 40: So are we using a separate Redis instance for this deduplication data?

src/deduplication.ts
```diff
@@ -0,0 +1,35 @@
...
+  try {
+    const res = await deduplicationRedisClient.setAsync(
+      `deduplication:${key}`,
+      'true',
+      'EX',
+      DEDUPE_WINDOW_SECONDS,
+      'NX'
+    );
+
+    if (res) {
```
tomerovadia: Just return `res`?

src/deduplication.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Implements redis-backed deduplication. The first time this is called with
+ * a given key, it will return true. After that, it will return false until
+ * the key expires in 1 hour.
+ *
+ * If there's an error talking to redis, it will return true (erring on the
+ * side of permitting duplication if there's a problem with Redis).
+ */
+export default async function deduplicate(key: string): Promise<boolean> {
```
tomerovadia: Can we name this something like `isUnique`? Was a bit hard to follow with this name

src/twilio_api_util.ts
```diff
@@ -29,6 +31,31 @@ export async function sendMessage(
     databaseMessageEntry.twilioSendTimestamp = new Date();
   }
 
+  if (
+    options.deduplicationId &&
+    !(await deduplicate(options.deduplicationId))
+  ) {
+    logger.warn(
+      `TWILIOAPIUTIL.sendMessage: Not sending duplicate Twilio message to ${options.userPhoneNumber}: ${message}`
```
tomerovadia: Can we add more info in this log (e.g. Slack thread and Slack message ts, if available from the databaseMessageEntry)?

tomerovadia (changes requested):
Took a quick look but have to jump on a call. Added a few comments but mostly looks great. Will take a final look after my call.

src/redis_client.ts
benweissmann: Optionally -- it will default to the main Redis instance. For staging and production I've set this to point to a different [logical database](https://redis.io/commands/select) on the same Redis server. So you can access these keys by connecting to the staging/prod redis server and then running `SELECT 1` (and `SELECT 0 ` switches you back to the main database)

tomerovadia: Sounds great

tomerovadia (approved):

tomerovadia (approved):

tomerovadia:
Tested this and it works. Very cool!

PR #106, opened 2020-10-10
Shortcut / Modal for opening + closing entrypoints

### LINKED ISSUE #102: Allow admins to give a channel N times as many voters as other channels in their entrypoint
I realized that we could easily give a channel N times as many voters as other channels in the same entry point by just adding it N times to the Redis entrypoint range. If `openPodsPullPacific` has `pacific-0` `pacific-0` `pacific-1` `pacific-2`, then pod 0 would get 50% of voters from that entry point and the other two would get 25% each. 

If that’s not too difficult of a follow-up after we get the basic open pods functionality up, perhaps this is something worth doing. The use case would be if we have a pod that has an especially high number of volunteers, or especially available volunteers, we can route them more voters than others.

fongandrew:
This implements a modal for opening / closing entrypoints for a given state. You can filter by state (or region) + demo vs non-demo. It looks like this.

Screen Shot 2020-10-09 at 2 00 52 PM

You'll need to create a new global shortcut with a callback ID of `open_close_channels`.

Screen Shot 2020-10-09 at 11 21 37 PM

Additional comments inline.

## Known Issues to Tackle in a Later PR
This PRs getting a bit long in the tooth, so I'd like to handle these in separate ones. We can delay rolling out access to the new modal by just not creating the shortcut on prod.

* We need some way to invalidate the Redis cache mapping channel names and IDs. Otherwise, newly created channels won't show up in this modal. We could listen on a Slack event for that, or maybe we could just add a button to manually invalidate the cache.

* ~We currently use a Redis list to track the channels open for any given entrypoint. This isn't ideal because it leaves open the possibility of a race condition in which a channel is added twice to the list (e.g. if two people check off something at once).~

* ~Need to post changes on channel opening / closing back to channel.~


package.json
```diff
@@ -15,6 +15,7 @@
     "dotenv": "^8.2.0",
     "express": "^4.17.1",
     "jshashes": "^1.0.8",
+    "lodash": "^4.17.20",
```
fongandrew, line 19: Adding this because I didn't want to reimplement stuff like `sortBy` and `uniq`.

src/async_jobs.ts
```diff
@@ -42,6 +55,52 @@ async function slackInteractivityHandler(
...
+      payload,
+      redisClient,
+      originatingSlackUserName,
+      viewId: getViewId(payload, interactivityMetadata),
+      action: payload.actions[0],
+    });
+    return;
+  }
+
   if (payload.type === 'block_actions' || payload.type === 'message_action') {
```
fongandrew: Something that might be nice for later is to simplify this function so it involves less nesting based on combinations of types and payloads and whatever. In theory, we should be able to do a giant switch statement on `action_id` and/or `callback_id`.

src/slack_api_util.ts
```diff
@@ -288,6 +288,20 @@ export async function updateSlackChannelNamesAndIdsInRedis(
...
+// Get Slack Channel Name / ID map, using cache if needed
+export async function getSlackChannelNamesAndIds(
+  redisClient: PromisifiedRedisClient
+): Promise<SlackChannelNamesAndIds | null> {
+  const slackChannelIds = await RedisApiUtil.getHash(
+    redisClient,
+    'slackPodChannelIds'
+  );
+  if (slackChannelIds) return slackChannelIds;
+  return updateSlackChannelNamesAndIdsInRedis(redisClient);
```
fongandrew, line 398: Helper function to query API directly if cache isn't set. Might be nice to use elsewhere since it eliminates needing to explicitly set the cache before grabbing stuff from it.

src/slack_block_util.ts
```diff
@@ -333,6 +354,170 @@ export function getErrorSlackView(
...
+          case PodUtil.ENTRYPOINT_TYPE.PUSH:
+            return pushOption;
+        }
+        logger.error(
+          `SLACKBLOCKUTIL.getOpenCloseModal: Invalid entrypoint type for ${channelName}: ${type}`
+        );
+      });
+
+      const accessory = {
+        type: 'checkboxes',
```
fongandrew: We could also switch this to the `multi_static_select` dropdown which is a little more space efficient than checkboxes, but it requires more clicks to actually enable / disable stuff.

src/slack_block_util.ts
```diff
@@ -412,45 +597,66 @@ export function replaceVoterPanelBlocks(
...
-
-            element.initial_option = newOption;
-            // Javascript modifies the blocks by reference, return success
-            return true;
-          }
-          if (element.type === 'users_select') {
-            element.initial_user = newInitialValue;
-            // Javascript modifies the blocks by reference, return success
-            return true;
-          }
```
fongandrew, line 448: This stuff just got moved below. It separates out finding an element by action ID from mutating it.

tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "lib": ["ESNext"],
     "module": "commonjs",
     "moduleResolution": "node",
-    "target": "ESNext",
+    "target": "ES2018",
```
fongandrew, line 7: I'm guess from the version of Node listed on `serverless.yml`, we're running Node v12? If so, we should switch the target to something not `ESNext`.

My motivation here was just to get optional chaining working (e.g. you can call `a?.b?.c` instead of `a && a.b && a.b.c`). If you specify `ESNext`, TypeScript will assume you're targeting the latest version of Node and since optional chaining is supported there, it won't transpile any optional chaining in your code. But optional chaining _isn't_ supported in Node v12. By telling TypeScript we're targeting ES2018, it'll convert the optional chaining to something that is supported on olde version of Node.

tomerovadia:
Thanks for doing this! I'm taking a look and will follow up with more a more detailed review, but a few quick thoughts:
1) Is there a way we can alert the user and maybe prevent them from moving forward if they try to close the last channel open in an entry point?
2) To your point about a Redis list, that structure helps us be able to do something like [this](https://github.com/tomerovadia/VoterHelpLine/issues/102). To your concern, can we check the list for if the value already exists, and respond accordingly if so?

fongandrew:
1. We could make submission a two-step process? So you check and uncheck everything, and when you submit, we check the entire set of boxes checked to determine if there are no channels open for an entry point.

2. Hmm. I don't think there's an atomic Redis command that lets you do that. So there's still a race condition. That might be fine I guess if there's a limited number of admins, but if we want to address it, we'd probably have to implement some kind of race condition check at the application level.

fongandrew:
Updated:
* Added a confirmation step if an entrypoint has no channels.
* Added support for different "weights" for channels.
* It now posts in the admin channel on change.
* Uses Redis transactions to avoid race condition.
* Regions are now separated from states in dropdown.

Screen Shot 2020-10-12 at 2 06 53 AM

Screen Shot 2020-10-12 at 2 03 08 AM

Screen Shot 2020-10-12 at 2 07 53 AM

Screen Shot 2020-10-12 at 2 08 38 AM

tomerovadia:
This is beautiful! I will review this evening. Thanks so much for doing this.

src/async_jobs.ts
```diff
@@ -29,6 +30,18 @@ import { UserInfo } from './types';
 
 export type InteractivityHandlerMetadata = { viewId?: string };
 
+function getViewId(
```
tomerovadia: Can you add a comment as to why the payload.view.id is preferred to the interactivityMetadata.viewId?

src/async_jobs.ts
```diff
@@ -42,11 +55,107 @@ async function slackInteractivityHandler(
...
+      redisClient,
+      originatingSlackUserName,
+      viewId: getViewId(payload, interactivityMetadata),
+      values: payload?.view?.state?.values,
+      action: payload.actions[0],
+    });
+    return;
+  }
+
+  // Handle open/close modal submission
```
tomerovadia: Can you add comments to clarify the difference between OPEN_CLOSE_CHANNELS_MODAL submission and OPEN_CLOSE_CHANNELS_MODAL_CONFIRM submission?

src/slack_interaction_handler.ts
```diff
@@ -297,6 +322,15 @@ export async function handleVolunteerUpdate({
   logger.info(
     `SLACKINTERACTIONHANDLER.handleVolunteerUpdate: Determined user interaction is a volunteer update`
   );
+  if (
+    !payload.actions ||
+    !payload.actions[0] ||
+    !payload.actions[0].selected_user
+  ) {
+    throw new Error(
+      'Expected seelcted_user in LACKINTERACTIONHANDLER.handleVolunteerUpdate'
```
tomerovadia: typo

src/slack_interaction_handler.ts
```diff
@@ -548,3 +582,164 @@ export async function handleResetDemo(
...
+  const isAdmin = await SlackApiUtil.isMemberOfAdminChannel(payload.user.id);
+  if (!isAdmin) {
+    logger.warn(
+      `SLACKINTERACTIONHANDLER.handleOpenCloseChannels: ${payload.user.id} is not an admin`
+    );
+    await SlackApiUtil.updateModal(
+      viewId,
+      SlackBlockUtil.getErrorSlackView(
+        SlackCallbackId.OPEN_CLOSE_CHANNELS_ERROR_MODAL,
+        'You must have access to the admin channel to do that'
```
tomerovadia: nit: can we say #admin-control-room, specifically? in case we add different admin channels later

src/slack_block_util.ts
```diff
@@ -333,6 +359,233 @@ export function getErrorSlackView(
   };
 }
 
+interface OpenCloseModalProps {
```
tomerovadia: Thoughts on moving all the open/close channel logic from slack_block_util to slack_block_entrypoint_util or something like that? Since it's ~doubling the size of the file.

src/slack_api_util.ts
```diff
@@ -288,6 +288,20 @@ export async function updateSlackChannelNamesAndIdsInRedis(
       slackChannelNamesAndIds
     );
   }
+
+  return slackChannelNamesAndIds;
+}
+
+// Get Slack Channel Name / ID map, using cache if needed
```
tomerovadia: should this be `populating cache if needed`? just wanna confirm my understanding

src/pod_util.ts
```diff
@@ -0,0 +1,240 @@
...
+  return sortBy(ret, 'channelName');
+}
+
+export async function getPodChannelState(
+  redisClient: PromisifiedRedisClient,
+  filters: PodFilters
+): Promise<{ pull: ChannelInfo[]; push: ChannelInfo[] }> {
+  // Get all channels from Redis cache or Slack API
+  const channelNamesAndIds =
+    (await SlackApiUtil.getSlackChannelNamesAndIds(redisClient)) || {};
```
tomerovadia, line 134: One thing to note is that the cache used in `getSlackChannelNamesAndIds` is currently only refreshed when a voter is routed to a channel that doesn't exist in the cache, either automatically from the lobby or manually via ROUTE_VOTER. The former creates a catch 22 (can't open the channel because it's not in the cache because a voter hasn't been routed there because the channel isn't open because it can't be opened because it's not on the modal because it's not in the cache etc...).

https://github.com/tomerovadia/VoterHelpLine/blob/96b3f90c3f01a53742b79705ebdf015325e4a819/src/router.ts#L512

Maybe you can modify `getSlackChannelNamesAndIds` to take a parameter that indicates whether the cache should be forced to update?

tomerovadia (commented):
Thanks for doing this! I haven't had the chance to review everything and will need more time, but in the meantime below and in-line in the code are some comments. It's very clear you took time to make sure every detail is carefully crafted and extra features (e.g. notices) are added, so thanks so much for all the time you've put into this.

1. Would you mind adding more comments? I think I totally underestimated the complexity of implementing something like this, and with the platform ramping up quickly and taking a lot of our attention, more comments might help us follow what's going on.
2. I tested the logic in my local environment and created a new channel to see if it was added to the modal. I ended up with demo channels mixed in with normal channels on the same page. Do you know how this might occur? I will look into it more as well.
Screen Shot 2020-10-12 at 11 51 05 PM

I will continue reviewing this tomorrow. Thanks again!

fongandrew:
Busy week at work, but I'll address comments sometime tomorrow.

tomerovadia:
> Busy week at work, but I'll address comments sometime tomorrow.

No worries! Things are taking off at VoteAmerica -- happy to update you if you're free for a quick call tomorrow, Friday or over the weekend.

src/async_jobs.ts
fongandrew: I'm going to remove this and just invoke directly in the handler code, but the original reason for this was sometimes we want a view ID from the interactivity metadata (e.g. a shortcut that opens a loading view and then updates it), and sometimes we're updating an existing one that's included in the payload (as with block actions).

src/pod_util.ts
fongandrew: I was thinking we have two ways to go about this:

1. Add a command to manually invalidate the cache as needed.
2. Listen for Slack events to invalidate or update the cache whenever a new channel is created.

I can add that in a follow up PR.

fongandrew:
OK, updated this PR to address comments and fix merge conflicts. With respect to the demo channels being mixed in the regular channels, I haven't been able to repro (it's possible resolving the merge conflicts fixed something). The contents of that message is derived from the view submission state payload though, so if it happens again, capturing that payload would be helpful.

fongandrew:
@tomerovadia OK, updated this PR to address comments and fix merge conflicts. With respect to the demo channels being mixed in the regular channels, I haven't been able to repro (it's possible resolving the merge conflicts fixed something). The contents of that message is derived from the view submission state payload though, so if it happens again, capturing that payload would be helpful.

tomerovadia:
I was just testing and got the following modal state when selecting the Pacific region. Do you know why demo-pacific-0 appears even though `Normal` is selected?
Screen Shot 2020-10-17 at 1 26 36 PM

tomerovadia:
> I was just testing and got the following modal state when selecting the Pacific region. Do you know why demo-pacific-0 appears even though `Normal` is selected?
> Screen Shot 2020-10-17 at 1 26 36 PM

Wait! I believe this is a bug with my local Redis. Please disregard (sorry for the confusion). I will follow up to confirm.

tomerovadia:
Sorry didn't mean to close this Pull Request.

external_config/slack.md
```diff
@@ -156,6 +156,14 @@ This should be added as a **global** shortcut.
 - **Short Description:** Show voters needing attention
 - **Callback ID:** show_needs_attention
 
+#### Open / Close Channels
+
+This should be added as a **global** shortcut.
+
+- **Name:** Open / Close Channels
```
tomerovadia: One minor suggestion: What are your thoughts on naming this `Manage Entry Points` (since you've included the weight functionality and an admin could change weights without completely opening/closing)?

tomerovadia (approved):

tomerovadia (approved):
All looks great!

PR #108, opened 2020-10-10
WIP support for outbound mms

benweissmann:
Started working on being able to send MMS from slack, but got a bit stuck on the API around getting file attachment -- would love some advice or pointers!


liewegas:
see https://github.com/tomerovadia/VoterHelpLine/pull/163

PR #109, opened 2020-10-10
Add support for receiving MMS

### LINKED ISSUE #89: Handle incoming MMS safely
E.g. what if a voter wants to text a photo of something going on at a polling place?

benweissmann:
Fixes #89

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -25,6 +25,7 @@ import {
 } from './async_jobs';
 import { handleTwilioStatusCallback } from './twilio_status_callback_handler';
 import { SlackInteractionEventPayload } from './slack_interaction_handler';
+import { TwilioRequestBody } from './twilio_util';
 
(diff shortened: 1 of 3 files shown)


src/app.ts
```diff
@@ -232,8 +234,8 @@ const handleIncomingTwilioMessage = async (
     return;
   }
 
-  const twilioPhoneNumber = req.body.To;
-  const userMessage = req.body.Body;
+  const twilioPhoneNumber = reqBody.To;
+  const userMessage = TwilioUtil.formatAttachments(reqBody);
```
tomerovadia, line 238: nit: Can we rename this (or add a comment) to make clearer that it returns a message? Something like `getMessageWithAttachments` or `formatMessageAttachments`?

tomerovadia (approved):

PR #111, opened 2020-10-11
Reorder states so we prioritize IN and OK below other states during parsing

benweissmann:
So we only categorize a state as Indiana (IN) or Oklahoma (OK) if they don't match any other states.


tomerovadia (approved):

tomerovadia:
Great idea thanks for doing this.

PR #115, opened 2020-10-12
Create sql/scripts and populate it with a voters_needing_attention script...

tomerovadia:
...and move migrations to sql/migrations.

This script shows voters who:
- have entered the Slack workspace
- do not have `REFUSED` or `SPAM` as their latest voter statuses, and
- EITHER:
      - have received no response from a volunteer, or
      - were not claimed by a volunteer.

It also:
- ranks voters in order of how long it has been since they messaged `helpline`, and displays that timestamp (in EST) and how long it has been
- displays the voter's most recent Slack channel (the Slack channel to which the voter's most recent message was sent).


tomerovadia:
There is one thing I want to double check about this query, and that might require making a change to logging.

tomerovadia:
@ynvovd feel free to ignore this PR but 1) the script might be useful as a starting point to you despite an issue I can discuss with you offline, and 2) I suggest changing the directory structure in the way you see in this PR so that you can check in and have your SQL scripts reviewed.

PR #116, opened 2020-10-12
Cleaner visuals

liewegas:
image

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -249,7 +249,7 @@ const introduceNewVoterToSlackChannel = async (
       `ROUTER.introduceNewVoterToSlackChannel: Passing voter message to Slack, slackChannelName: ${slackChannelName}, parentMessageTs: ${response.data.ts}.`
     );
     await SlackApiUtil.sendMessage(
-      `*${userInfo.userId.substring(0, 5)}:* ${userMessage}`,
+      `${userMessage}`,
       { parentMessageTs: response.data.ts, channel: response.data.channel },
       inboundDbMessageEntry,
       userInfo
@@ -752,7 +752,7 @@ export async function determineVoterState(
     `ROUTER.determineVoterState: Passing voter message to Slack, slackChannelName: ${lobbyChannelId}, parentMessageTs: ${lobbyParentMessageTs}.`
   );
   await SlackApiUtil.sendMessage(
-    `*${userId.substring(0, 5)}:* ${userMessage}`,
+    `${userMessage}`,
     { parentMessageTs: lobbyParentMessageTs, channel: lobbyChannelId },
     inboundDbMessageEntry,
     userInfo
@@ -913,7 +913,7 @@ export async function handleDisclaimer(
   userInfo.lastVoterMessageSecsFromEpoch = Math.round(Date.now() / 1000);
 
   await SlackApiUtil.sendMessage(
-    `*${userId.substring(0, 5)}:* ${userMessage}`,
+    `${userMessage}`,
     slackLobbyMessageParams,
     inboundDbMessageEntry,
     userInfo
--- a/src/router.ts
+++ b/src/router.ts
@@ -980,7 +980,7 @@ export async function handleClearedVoter(
   userInfo.lastVoterMessageSecsFromEpoch = nowSecondsEpoch;
 
   await SlackApiUtil.sendMessage(
-    `*${userId.substring(0, 5)}:* ${userOptions.userMessage}`,
+    `${userOptions.userMessage}`,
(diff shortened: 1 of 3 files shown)


src/router.ts
```diff
@@ -980,7 +980,7 @@ export async function handleClearedVoter(
   userInfo.lastVoterMessageSecsFromEpoch = nowSecondsEpoch;
 
   await SlackApiUtil.sendMessage(
-    `*${userId.substring(0, 5)}:* ${userOptions.userMessage}`,
+    `${userOptions.userMessage}`,
```
tomerovadia, line 983: This is only one of the points at which messages from the voter are relayed to Slack, so you'll have to remove the prefix for all, if that's what you intended. See line 916 for example.

src/slack_message_formatter.ts
```diff
@@ -27,12 +17,23 @@ export function formatMessageHistory(
...
-    );
+    const specialSlackTimestamp = `<!date^${timeSinceEpochSecs}^{time} {date_short}|${messageObject.timestamp}>`;
+    if (messageObject.direction == 'INBOUND') {
+      return (
+        `:bust_in_silhouette: *Voter ${userId}*  ` +
+        specialSlackTimestamp +
+        '\n' +
+        formatMessageBlock(messageObject.message, '*')
+      );
+    } else {
```
tomerovadia: I love the gear icon and the bust_in_silhouette icons! One other thing to keep in mind is that voters can be routed manually to a new channel, in which case the conversation history is similarly carried over to the new channel, including messages from other volunteers. Can we have a separate icon for volunteer messages? Maybe a different bust_in_silhouette color?

Here is an example: https://voteamericahelpline.slack.com/archives/G01BQE0THSA/p1602363906001000

You can test this locally by going to the `#admin-control-room` and typing `@${yourbotuser} ROUTE_VOTER ${userid} ${phonenumber with leading "+1"} ${destinationSlackChannel without leading #}`. Also, please also test routing a voter to a new channel, having additional conversation in that channel, and then routing them back to their original channel, which should post in their original thread in the original channel, adding only the incremental message history that was missed.

tomerovadia (changes requested):
This is beautiful! Got the opinions of a few others and they're extremely excited as well. I just have a few minors comments and then if you can merge this in tonight I can include it in a release tonight.

One high-level thought: is there a way we can make clear that the (singular) messages posted by the bot are coming from the voter? That's what the userId prefix was intended to do, but I can see an argument that this quickly becomes clear to a new volunteer. This might be a question for Brenna/Fuzzy, but maybe we should rename the bot (which we can do in the app settings), or change its avatar? The other thing to consider is that the bot does more than just relay voter messages -- it also posts when a voter status or volunteer is updated, or when a voter is routed. So we need to make sure those are differentiated from voter messages.

src/router.ts
liewegas: > One high-level thought: is there a way we can make clear that the (singular) messages posted by the bot are coming from the voter? That's what the userId prefix was intended to do, but I can see an argument that this quickly becomes clear to a new volunteer. This might be a question for Brenna/Fuzzy, but maybe we should rename the bot (which we can do in the app settings), or change its avatar? The other thing to consider is that the bot does more than just relay voter messages -- it also posts when a voter status or volunteer is updated, or when a voter is routed. So we need to make sure those are differentiated from voter messages.

Yeah.  The bot-related messages still have the bold **Helpline** prefix on them... I was hoping that would be sufficiently clear (since voter's can bold their stuff)?  At the end of the day the bot only has one avatar, so we have to share somehow.

tomerovadia: Sounds good to me (I think you mean the **Operator:** prefix, right?). I think it'll be clear to volunteers what's going on, so this sounds good.

liewegas:
Updated!  Used the :adult::skin-tone-4: emoji since that was the most neutral one I could find that was distinct from a silhouette

tomerovadia:
Can you fix the tests?

src/slack_message_formatter.ts
```diff
@@ -34,7 +34,7 @@ export function formatMessageHistory(
       );
     } else {
       return (
-        ':adult: *Helpline*  ' +
+        ':adult::skin-tone-4: *Helpline*  ' +
```
tomerovadia: Thoughts on using the emoji default yellow?

liewegas: sure!

tomerovadia:
All looks great! Feel free to merge.

tomerovadia (approved):

PR #122, opened 2020-10-13
Track thread needs_attention state; shortcut to view threads needing attention

liewegas:
- new table `threads` tracks needs_attention state
- new shortcut shows threads for the current user needing attention
- sql statement to create the table after the fact, and populate it with historic data.  (Note that this doesn't account for #lobby and #demo-lobby... those should be manually set to needs_attention=false.)
- message shortcuts to set/clear needs_attention

TODO
- [ ] daily reminder
- [x] slash command to summarize needs_attention threads in a channel
- [x] operator comamnds


sql/20201012.sql
```diff
@@ -0,0 +1,36 @@
...
+        , row_number() OVER (PARTITION BY slack_parent_message_ts ORDER BY COALESCE(slack_send_timestamp, slack_receive_timestamp) DESC) as rn
+    FROM messages
+    WHERE slack_parent_message_ts IS NOT NULL AND slack_channel IS NOT NULL
+)
+SELECT
+    slack_parent_message_ts
+    , channel_id
+    , user_id
+    , user_phone_number
+    , CASE WHEN direction='INBOUND' THEN true ELSE false END as needs_attention
```
benweissmann: Unassigned threads should be marked as needs attention, even if the last message was outbound

sql/20201012.sql
```diff
@@ -0,0 +1,36 @@
+--- Table to explicit track thread state
+
+CREATE TABLE threads (
+    slack_parent_message_ts text,
```
benweissmann, line 4: You should add indexes to all of these columns (except maybe user_phone_number?)

src/db_api_util.ts
```diff
@@ -591,6 +756,22 @@ export async function logVolunteerVoterClaimToDb(
...
 }
 
+export async function getVoterHasVolunteer(
+  userId: string,
+  userPhoneNumber: string
+): Promise<boolean> {
+  const client = await pool.connect();
+  try {
+    const result = await client.query(
+      'SELECT * FROM volunteer_voter_claims WHERE user_id = $1 AND user_phone_number = $2',
```
benweissmann: This is a bit faster: `SELECT EXISTS(SELECT 1 FROM volunteer_voter_claims WHERE user_id = $1 AND user_phone_number = $2))` (because it's not reading any values, it can figure out the answer straight from the index without having to load in the pages of data)

benweissmann (commented):
This looks amazing!

external_config/slack.md
```diff
@@ -81,3 +81,24 @@ This should be added as a **messages** shortcut.
 - **Name:** Reset Demo
 - **Short Description:** Closes this conversation, allowing the volunteer to message the demo line again as a new mock voter.
 - **Callback ID:** reset_demo
+
+##### Add to ToDo
+
+- Messages shortcut
```
tomerovadia: To keep the format consistent, here and below can you match the line in `Reset Demo` (or change it to match what you have here)?

`This should be added as a **messages** shortcut.`

sql/20201012.sql
```diff
@@ -0,0 +1,36 @@
...
+    channel_id text,
```
tomerovadia: For consistency across tables can you name this `slack_channel_id`?

sql/20201012.sql
```diff
@@ -0,0 +1,36 @@
...
+    user_id text,
+	user_phone_number text,
```
tomerovadia: unindent

sql/20201012.sql
```diff
@@ -0,0 +1,36 @@
...
+    needs_attention bool,
+    updated_at timestamp without time zone,
+    PRIMARY KEY (slack_parent_message_ts)
```
tomerovadia: I believe you'll want `slack_parent_message_ts` and `slack_channel_id` to both be a multi-column primary key, since `slack_parent_message_ts` is only guaranteed to be unique within a certain channel.

sql/20201012.sql
```diff
@@ -0,0 +1,36 @@
...
+);
+
+INSERT INTO threads
+WITH foo AS (
+    SELECT
+        slack_parent_message_ts
+        , slack_channel as channel_id
```
tomerovadia: Are there formatting issues with the commas?

sql/20201012.sql
```diff
@@ -0,0 +1,36 @@
...
+    slack_parent_message_ts text,
+    channel_id text,
+    user_id text,
+	user_phone_number text,
+    needs_attention bool,
+    updated_at timestamp without time zone,
+    PRIMARY KEY (slack_parent_message_ts)
+);
+
+INSERT INTO threads
```
tomerovadia, line 22: can you add a comment about what this script does?

src/app.ts
```diff
@@ -607,8 +607,9 @@ app.post(
     const metadata: InteractivityHandlerMetadata = {};
 
     if (
-      payload.type === 'message_action' &&
-      payload.callback_id === 'reset_demo'
+      (payload.type === 'message_action' &&
+        payload.callback_id === 'reset_demo') ||
+      payload.type === 'shortcut'
```
tomerovadia, line 659: indent

sql/20201012.sql
```diff
@@ -0,0 +1,36 @@
...
+        , COALESCE(slack_send_timestamp, slack_receive_timestamp) as updated_at
+        , direction
+        , automated
+        , row_number() OVER (PARTITION BY slack_parent_message_ts ORDER BY COALESCE(slack_send_timestamp, slack_receive_timestamp) DESC) as rn
+    FROM messages
+    WHERE slack_parent_message_ts IS NOT NULL AND slack_channel IS NOT NULL
+)
+SELECT
+    slack_parent_message_ts
+    , channel_id
```
tomerovadia: slack_channel_id for consistency with other tables

src/async_jobs.ts
```diff
@@ -42,6 +42,25 @@ async function slackInteractivityHandler(
     );
   }
 
+  if (payload.type === 'shortcut') {
```
tomerovadia, line 83: can you add a comment that this is for global shortcuts?

src/async_jobs.ts
```diff
@@ -42,6 +62,25 @@ async function slackInteractivityHandler(
...
+    const { viewId } = interactivityMetadata;
+    if (!viewId) {
+      throw new Error(
+        'slackInteractivityHandler called for message_action without viewId'
+      );
+    }
+    switch (payload.callback_id) {
+      case 'show_needs_attention': {
+        await SlackInteractionHandler.receiveShowNeedsAttention({
```
tomerovadia: can you rename to `handleShowNeedsAttention`? Elsewhere I use `receive` to mean receive the command but don't execute it (ask for confirmation), and `handle` is to do the actual execution logic

src/async_jobs.ts
```diff
@@ -102,6 +141,21 @@ async function slackInteractivityHandler(
           `SERVER POST /slack-interactivity: Determined user interaction is a message shortcut.`
         );
 
+        if (payload.callback_id === 'set_needs_attention') {
```
tomerovadia: can you make this a switch statement on the callback_id (and I believe the below logic is callback_id = reset_demo)?

src/db_api_util.ts
```diff
@@ -86,6 +86,184 @@ export type DatabaseCommandEntry = {
...
+
+export type ThreadInfo = {
+  slackParentMessageTs: string;
+  channelId: string;
+  userId: string | null;
+  userPhoneNumber: string | null;
+  age: number | null;
+};
+
+export async function newThreadToDb(
```
tomerovadia: can we make this `logThreadToDb` to keep the naming consistent across the file?

src/db_api_util.ts
```diff
@@ -86,6 +86,184 @@ export type DatabaseCommandEntry = {
...
+      'INSERT INTO threads (slack_parent_message_ts, channel_id, user_id, user_phone_number, needs_attention, updated_at) VALUES ($1, $2, $3, $4, $5, NOW())',
+      [
+        databaseThreadEntry.slackParentMessageTs,
+        databaseThreadEntry.channelId,
+        databaseThreadEntry.userId,
+        databaseThreadEntry.userPhoneNumber,
+        databaseThreadEntry.needsAttention,
+      ]
+    );
+    logger.info('DBAPIUTIL.newThreadToDb: Successfully created thread');
```
tomerovadia: logThreadToDb

src/db_api_util.ts
```diff
@@ -86,6 +86,184 @@ export type DatabaseCommandEntry = {
   failureReason?: string | null;
 };
 
+export type DatabaseThreadEntry = {
```
tomerovadia, line 89: I have this file ordered such that the functions associated with tables that are most important (e.g. messages) are at top. Can you move these functions to the bottom?

src/db_api_util.ts
```diff
@@ -86,6 +86,184 @@ export type DatabaseCommandEntry = {
...
+}
+
+export async function setThreadNeedsAttentionToDb(
+  slackParentMessageTs: string,
+  needsAttention: boolean
+): Promise<void> {
+  const client = await pool.connect();
+  try {
+    await client.query(
+      'UPDATE threads SET needs_attention = $1 WHERE slack_parent_message_ts = $2;',
```
tomerovadia: slack_parent_message_ts is only guaranteed to be unique within a channel, so we need to add channel to the WHERE clause here

src/db_api_util.ts
```diff
@@ -86,6 +86,184 @@ export type DatabaseCommandEntry = {
...
+      `DBAPIUTIL.setThreadNeedsAttentionToDb: Set thread ${slackParentMessageTs} needs_attention=${needsAttention}`
+    );
+  } catch (error) {
+    logger.info('Failed to update threads; ignoring for now!');
+  } finally {
+    client.release();
+  }
+}
+
+export async function getThreadLatestMessage(
```
tomerovadia: This function only gets the message ts, right? Can we rename `getThreadLatestMessageTs`?

src/db_api_util.ts
```diff
@@ -86,6 +86,184 @@ export type DatabaseCommandEntry = {
...
+  try {
+    let result = null;
+    if (channelId) {
+      result = await client.query(
+        `SELECT
+         t.slack_parent_message_ts as id, t.channel_id, t.user_phone_number, t.user_id, EXTRACT(EPOCH FROM now() - t.updated_at) as age
+        FROM threads t
+        WHERE t.needs_attention AND t.channel_id = $1
+          AND NOT EXISTS (SELECT FROM volunteer_voter_claims c WHERE t.user_id=c.user_id
+          AND t.user_phone_number=c.user_phone_number)`,
```
tomerovadia: can you indent this to make it clear it's part of the subquery?

src/db_api_util.ts
```diff
@@ -86,6 +86,184 @@ export type DatabaseCommandEntry = {
...
+  const client = await pool.connect();
+  try {
+    let result = null;
+    if (channelId) {
+      result = await client.query(
+        `SELECT
+         t.slack_parent_message_ts as id, t.channel_id, t.user_phone_number, t.user_id, EXTRACT(EPOCH FROM now() - t.updated_at) as age
+        FROM threads t
+        WHERE t.needs_attention AND t.channel_id = $1
+          AND NOT EXISTS (SELECT FROM volunteer_voter_claims c WHERE t.user_id=c.user_id
```
tomerovadia: you shouldn't need to check both the user_id and user_phone_number -- the user_id is derived from the user_phone_number

src/db_api_util.ts
```diff
@@ -86,6 +86,184 @@ export type DatabaseCommandEntry = {
...
+export async function getUnclaimedVoters(
+  channelId: string | null
+): Promise<ThreadInfo[]> {
+  const client = await pool.connect();
+  try {
+    let result = null;
+    if (channelId) {
+      result = await client.query(
+        `SELECT
+         t.slack_parent_message_ts as id, t.channel_id, t.user_phone_number, t.user_id, EXTRACT(EPOCH FROM now() - t.updated_at) as age
```
tomerovadia: is age the age of the voter or age of the todo? can you make that clearer?

src/db_api_util.ts
```diff
@@ -86,6 +86,184 @@ export type DatabaseCommandEntry = {
...
+    }));
+  } catch (error) {
+    logger.info('Failed to query threads; ignoring for now!');
+    return [];
+  } finally {
+    client.release();
+  }
+}
+
+export async function getThreadNeedsAttentionFor(
```
tomerovadia: I see the difference between getThreadNeedsAttentionFor and getThreadsNeedingAttentionFor but can you please add comments since it can be confusing?

src/db_api_util.ts
```diff
@@ -86,6 +86,184 @@ export type DatabaseCommandEntry = {
...
+): Promise<boolean> {
+  logger.info(`ENTERING DBAPIUTIL.getThreadNeedsAttentionFor`);
+  logger.info(
+    `DBAPIUTIL.getThreadNeedsAttentionFor: Looking up thread:${slackParentMessageTs}`
+  );
+
+  const client = await pool.connect();
+  try {
+    const result = await client.query(
+      'SELECT needs_attention FROM threads WHERE slack_parent_message_ts = $1',
```
tomerovadia: You're going to need to filter by both slack_parent_message_ts and slack_channel_id to guarantee you only get one row

src/db_api_util.ts
```diff
@@ -123,10 +301,37 @@ export async function logMessageToDb(
         databaseMessageEntry.archived,
       ]
     );
-
     logger.info(
       `DBAPIUTIL.logMessageToDb: Successfully inserted message into PostgreSQL database.`
     );
+
```
tomerovadia: I'd prefer to keep logMessageToDb's responsibility strictly to logging a message to the DB. Can we add this logic elsewhere, perhaps in a function called alongside where logMessageToDb is called?

src/db_api_util.ts
```diff
@@ -591,6 +796,22 @@ export async function logVolunteerVoterClaimToDb(
...
 }
 
+export async function getVoterHasVolunteer(
+  userId: string,
+  userPhoneNumber: string
+): Promise<boolean> {
+  const client = await pool.connect();
+  try {
+    const result = await client.query(
+      'SELECT * FROM volunteer_voter_claims WHERE user_id = $1 AND user_phone_number = $2',
```
tomerovadia: nit: you shouldn't need to check both user_id and user_phone_number. doesn't hurt if you prefer it though. if you choose one go with user_id

src/router.ts
```diff
@@ -704,6 +731,11 @@ const routeVoterToSlackChannel = async (
     `timestampOfLastMessageInThread: ${timestampOfLastMessageInThread}`
   );
 
+  await DbApiUtil.setThreadNeedsAttentionToDb(
```
tomerovadia, line 754: This updates the thread to which a voter is returning to have the needsAttention status of the thread from which they are coming from, right? Can you add a comment to that effect?

src/router.ts
```diff
@@ -632,6 +650,15 @@ const routeVoterToSlackChannel = async (
     // Remember the voter's thread in this channel.
     userInfo[response.data.channel] = response.data.ts;
 
+    // Create the thread
```
tomerovadia: can you add to the comment that this new thread will adopt the old thread's needAttention state?

src/slack_api_util.ts
```diff
@@ -168,6 +203,64 @@ export function copyUserInfoToDbMessageEntry(
     userInfo.lastVoterMessageSecsFromEpoch;
 }
 
+export async function fetchSlackChannelMap(): Promise<
+  [Record<string, string>, Record<string, string>]
+> {
+  const response = await slackAPI.get('conversations.list', {
+    params: {
+      limit: 1000,
```
tomerovadia: Slack recommends no more than 200 at a time. We have a function fetchSlackChannelNamesAndIds in this file that already does this with pagination https://api.slack.com/methods/conversations.list

src/slack_api_util.ts
```diff
@@ -168,6 +203,64 @@ export function copyUserInfoToDbMessageEntry(
...
+    logger.error(
+      `SLACKAPIUTIL.fetchSlackChannelId: Failed to list Slack channels. Error: ${JSON.stringify(
+        response.data
+      )}.`
+    );
+  }
+  return [fw, bw];
+}
+
+export async function fetchSlackChannelId(
```
tomerovadia: We have a Redis cache to look up a Slack id given a Slack name: RedisApiUtil.slackPodChannelIds

src/slack_api_util.ts
```diff
@@ -26,13 +26,48 @@ type SlackSendMessageResponse = {
...
+  parse?: boolean;
+  unfurl_links?: boolean;
   blocks?: SlackBlock[];
 };
 
 type SlackChannelNamesAndIds = {
   [channelId: string]: string; // mapping of channel ID to channel name
 };
 
+export async function getThreadPermalink(
```
tomerovadia, line 39: Can we have the SlackApiUtil function just do the Slack API call and have the DB lookup separate?

src/app.ts
liewegas: this is what prettier did

sql/20201012.sql
```diff
@@ -20,6 +25,7 @@ WITH foo AS (
         , COALESCE(slack_send_timestamp, slack_receive_timestamp) as updated_at
         , direction
         , automated
+        , archived
```
tomerovadia: Is `archived` used?

sql/20201012.sql
```diff
@@ -18,24 +18,28 @@ CREATE INDEX ON threads (updated_at);
 INSERT INTO threads
 WITH foo AS (
     SELECT
-        slack_parent_message_ts
+        m.slack_parent_message_ts
         , slack_channel as channel_id
```
tomerovadia: can you name this `slack_channel_id` for consistency?

sql/20201012.sql
```diff
@@ -18,24 +18,28 @@ CREATE INDEX ON threads (updated_at);
...
+    )
+    WHERE m.slack_parent_message_ts IS NOT NULL AND slack_channel IS NOT NULL
 )
 SELECT
     slack_parent_message_ts
     , channel_id
     , user_id
     , user_phone_number
-    , CASE WHEN direction='INBOUND' THEN true ELSE false END as needs_attention
+    , CASE WHEN direction='INBOUND' OR volunteer_slack_user_id IS NULL) THEN true ELSE false END as needs_attention
```
tomerovadia: Am I interpreting this correctly that this defines `needs_attention` as either a) the last message was from the voter, or b) no volunteer has claimed the voter?

If so, can you add a comment to that effect?

src/db_api_util.ts
```diff
@@ -212,12 +212,12 @@ export async function getThreadsNeedingAttentionFor(
...
       `SELECT
-         t.slack_parent_message_ts as id, t.channel_id, t.user_phone_number, t.user_id, EXTRACT(EPOCH FROM now() - t.updated_at) as age
-        FROM threads t, volunteer_voter_claims c
-        WHERE t.needs_attention
-          AND t.user_id=c.user_id
-          AND t.user_phone_number=c.user_phone_number
-          AND c.volunteer_slack_user_id=$1`,
+         slack_parent_message_ts as id, channel_id, user_phone_number, user_id, EXTRACT(EPOCH FROM now() - updated_at) as age
+        FROM threads t
+        WHERE needs_attention AND EXISTS (
```
tomerovadia: This is checking whether the user has ever claimed the voter. Did you intend to check whether the voter is currently claimed by the user? If so you'll need to get the latest voter_volunteer_claim to see the latest volunteer who claimed the voter.

src/db_api_util.ts
```diff
@@ -145,19 +145,44 @@ export async function setThreadNeedsAttentionToDb(
...
+    );
+  } catch (error) {
+    logger.info('Failed to update threads; ignoring for now!');
+  } finally {
+    client.release();
+  }
+}
+
 export async function getThreadLatestMessage(
   slackParentMessageTs: string
```
tomerovadia: Is the purpose of `history_ts` to more easily be able to access the latest message in a thread? If so, why do you need to join on messages here -- cant you use `history_ts` instead? Looks like you use both, but does that defeat the purpose of `history_ts`?

external_config/slack.md
```diff
@@ -92,23 +92,26 @@ This should be added as a **messages** shortcut.
...
+
+- **Name:** Clear Needs Attention
+- **Short Description**: Remove this thread from your Needs Attention list
 - **Callback ID**: clear_needs_attention
 
 ##### ToDo
 
-- Global shortcut
-- **Name:** ToDo
+This should be added as a **Global** shortcut.
```
tomerovadia: lowercase `global` (or uppercase `messages` above)

src/slack_interaction_handler.ts
```diff
@@ -385,23 +386,27 @@ export async function handleCommandUnclaimed(
...
-    [channelFw, channelBw] = await SlackApiUtil.fetchSlackChannelMap();
+  const slackChannelIds = text
+    ? await RedisApiUtil.getHash(redisClient, 'slackPodChannelIds')
+    : {};
+  let slackChannelNames: Record<string, string> = {};
+  for (const name in slackChannelIds) {
+    slackChannelNames[slackChannelIds[name]] = name;
   }
+
+  logger.info(JSON.stringify(slackChannelIds));
```
tomerovadia: This can be an extremely long list. Can we omit, or if you think it's necessary, add some context so when we see it in the logs we know what it's indicating?

src/slack_interaction_handler.ts
```diff
@@ -385,23 +386,27 @@ export async function handleCommandUnclaimed(
 ): Promise<void> {
   const outputChannelId = channelId;
 
-  let channelFw: Record<string, string> = {};
-  let channelBw: Record<string, string> = {};
-  if (text) {
-    [channelFw, channelBw] = await SlackApiUtil.fetchSlackChannelMap();
+  const slackChannelIds = text
```
tomerovadia: If you're using this to get the Slack channel ID from the name, could the slash command escaping option take care of that instead?

src/slack_interaction_handler.ts
```diff
@@ -417,9 +417,12 @@ export async function handleCommandUnclaimed(
       (text === '*' ? 'all channels' : `#${channelName}`),
   ];
   for (const x of threads) {
```
tomerovadia: Can we replace `x` with `thread` so this block is easier to read?

src/async_jobs.ts
```diff
@@ -42,66 +78,59 @@ async function slackInteractivityHandler(
...
-        logger.debug(
-          `SERVER POST /slack-interactivity: Received an interaction for a voter who no longer exists in Redis.`
+    logger.info(
+      `SERVER POST /slack-interactivity: Determined user interaction is a message shortcut.`
+    );
+
+    switch (payload.callback_id) {
+      case 'set_needs_attention': {
+        await DbApiUtil.setThreadNeedsAttentionToDb(
+          payload.message.thread_ts || payload.message.ts,
```
tomerovadia, line 119: Is there a reason one of these would exist and the other wouldn't? If so, can you add a comment?

src/async_jobs.ts
```diff
@@ -42,66 +78,59 @@ async function slackInteractivityHandler(
...
+  }
+
+  // Message shortcut
+  if (payload.type === 'message_action') {
     const originatingSlackChannelName = await SlackApiUtil.fetchSlackChannelName(
       payload.channel.id
     );
 
     const redisHashKey = `${payload.channel.id}:${payload.message.ts}`;
     const redisData = await RedisApiUtil.getHash(redisClient, redisHashKey);
```
tomerovadia, line 109: Do you handle a case in which this message command is used on a message that is not part of a still-active voter thread (e.g. no redisData is returned)?

src/async_jobs.ts
```diff
@@ -126,38 +155,77 @@ async function slackInteractivityHandler(
           "actionTs": payload.action_ts
         } as SlackModalPrivateMetadata;
 
-        if (!originatingSlackChannelName || !redisData) {
```
tomerovadia: Do you drop this logic entirely? After your logic, does a feedback modal still appear if a user attempts Reset Demo in an invalid context, or does the loading modal hang in these situations?

external_config/slack.md
```diff
@@ -81,6 +81,13 @@ Important: The Slack bot must be given access to each of these channels.
 - *Short Description:* List unclaimed voters in a channel
```
tomerovadia: Is `Slash Commands` a section under `Interactivity & Shortcuts`? It appears to me as a section of its own on the left.

external_config/slack.md
```diff
@@ -81,6 +81,13 @@ Important: The Slack bot must be given access to each of these channels.
...
 - *Usage Hint:* optional: pass channel name, or * for all channels
 
+##### Needs attention
+
+- *Command:* /needs-attention
+- *Request URL:* base URL + /slack-command
+- *Short Description:* List voters needing your attention
+- *Usage Hint:*
```
tomerovadia: Can you add a usage hint?

src/slack_interaction_handler.ts
```diff
@@ -457,16 +458,8 @@ export async function handleCommandUnclaimed(
...
-  viewId: string;
-}): Promise<void> {
-  logger.info(`Entering SLACKINTERACTIONHANDLER.handleShowNeedsAttention`);
-  const threads =
-    (await DbApiUtil.getThreadsNeedingAttentionFor(payload.user.id)) || [];
+async function getNeedsAttentionList(userId: string): Promise<string[]> {
+  const threads = (await DbApiUtil.getThreadsNeedingAttentionFor(userId)) || [];
 
   const urls: string[] = [];
   for (const x of threads) {
```
tomerovadia: Can you use `thread` instead of `x`?

src/slack_interaction_handler.ts
```diff
@@ -479,20 +472,64 @@ export async function handleShowNeedsAttention({
     urls.push(url);
   }
 
+  const lines = threads.map(
```
tomerovadia, line 491: Can you remove this use of `.map` and incorporate the logic into the above loop through threads? Especially because map is asynchronous it's not great to have it deconstruct a separate data structure as a side effect.

src/slack_interaction_handler.ts
```diff
@@ -479,20 +472,64 @@ export async function handleShowNeedsAttention({
...
+          text:
+            `*${lines.length}* voters need attention from *${userName}*\n` +
+            lines.join('\n'),
+        },
+      },
+    ],
+  });
+}
+
+export async function handleShowNeedsAttention({
```
tomerovadia: Is the difference between `handleCommandNeedsAttention` and `handleShowNeedsAttention` that the former is a slash command and the latter is a global shortcut? If so, can we rename the latter to `handleShortcutNeedsAttention`?

src/slack_interaction_handler.ts
```diff
@@ -396,26 +402,26 @@ export async function handleCommandUnclaimed(
...
   logger.info(JSON.stringify(slackChannelIds));
-  if (text && text != '*') {
-    if (text[0] == '#') {
-      text = text.substr(1);
+  if (arg && arg != '*') {
+    if (arg[0] == '#') {
+      arg = arg.substr(1);
     }
-    if (!(text in slackChannelIds)) {
+    if (!(arg in slackChannelIds)) {
```
tomerovadia, line 410: Can we replace much of this logic with the Slack slash command mention escaping feature?

src/slack_interaction_handler.ts
```diff
@@ -396,26 +402,26 @@ export async function handleCommandUnclaimed(
...
-    channelName = text;
+    channelName = arg;
     channelId = slackChannelIds[channelName];
   }
 
   const threads = await DbApiUtil.getUnclaimedVoters(
-    text === '*' ? null : channelId
+    arg === '*' ? null : channelId
   );
   const lines: string[] = [
```
tomerovadia, line 423: Am I understanding correctly that these initial values are of information that appears at the top of the message sent in response to the slash command? If so can you add a comment?

src/slack_interaction_handler.ts
```diff
@@ -383,11 +383,17 @@ export function prettyTimeInterval(seconds: number): string {
 export async function handleCommandUnclaimed(
   channelId: string,
   channelName: string,
+  userId: string,
   text: string
 ): Promise<void> {
   const outputChannelId = channelId;
 
-  const slackChannelIds = text
+  var arg = text;
```
tomerovadia: Can you rename arg (perhaps `slashCommandArg`)? Its purpose wasn't clear to me.

src/slack_interaction_handler.ts
```diff
@@ -396,26 +402,26 @@ export async function handleCommandUnclaimed(
   }
 
   logger.info(JSON.stringify(slackChannelIds));
-  if (text && text != '*') {
-    if (text[0] == '#') {
-      text = text.substr(1);
+  if (arg && arg != '*') {
+    if (arg[0] == '#') {
```
tomerovadia, line 407: `===` instead of `==`

tomerovadia, line 407: I see what you're doing here but can you please add comments so that other readers or us in the future can quickly discern why this is necessary?

src/slack_interaction_handler.ts
```diff
@@ -396,26 +402,26 @@ export async function handleCommandUnclaimed(
...
-    if (!(text in slackChannelIds)) {
+    if (!(arg in slackChannelIds)) {
       await SlackApiUtil.sendMessage(`Channel #${channelName} not found`, {
         channel: outputChannelId,
       });
       return;
     }
-    channelName = text;
+    channelName = arg;
     channelId = slackChannelIds[channelName];
```
tomerovadia: Am I understanding correctly that you're reassigning the original `channelId` here, which was the channel in which the slash command was issued, to now be the channel that was indicated in the slash command? If so, can you use separate variables for those two separate concepts?

Or are you intending to have the slash command respond the same way to `*` and it would to being invoked without an argument? If so, can you please refactor or add comments? I'm having a hard time following what's going on.

src/slack_interaction_handler.ts
```diff
@@ -396,26 +402,26 @@ export async function handleCommandUnclaimed(
   }
 
   logger.info(JSON.stringify(slackChannelIds));
-  if (text && text != '*') {
-    if (text[0] == '#') {
-      text = text.substr(1);
+  if (arg && arg != '*') {
```
tomerovadia, line 406: What happens if arg is empty string? AFAICT you call `getUnclaimedVoters(null)` similarly to if it were `*` but later it appears the code would handle it as if it were a state.

src/slack_interaction_handler.ts
```diff
@@ -383,11 +383,17 @@ export function prettyTimeInterval(seconds: number): string {
...
   channelId: string,
   channelName: string,
+  userId: string,
   text: string
 ): Promise<void> {
   const outputChannelId = channelId;
 
-  const slackChannelIds = text
+  var arg = text;
+  if (text && !SlackApiUtil.isMemberOfAdminChannel(userId)) {
```
tomerovadia, line 391: I'm having hard time following what you intend to happen if arg is an empty string. Can you send a message letting the user know that this slash command is for admins only?

src/slack_interaction_handler.ts
```diff
@@ -383,11 +383,17 @@ export function prettyTimeInterval(seconds: number): string {
 export async function handleCommandUnclaimed(
```
tomerovadia, line 382: I believe there is something specific you need to do in order to ensure that the slash command itself appears (before the response). Did you come across that / does the flow of how this works currently include the slash command itself, so that it's clear to other podmates why the message with the data was sent?

src/db_api_util.ts
```diff
@@ -697,18 +697,24 @@ export async function getThreadsNeedingAttentionFor(
...
+        FROM volunteer_voter_claims
+      )
+      SELECT
         slack_parent_message_ts
         , slack_channel_id
-        , user_id
+        , t.user_id
         , EXTRACT(EPOCH FROM now() - updated_at) as age
-        FROM threads t
+        FROM threads t, claims c
```
tomerovadia, line 873: I believe this crossjoin will duplicate every row in threads by every row in claims which can cause issues and I don't believe is what you're looking for here (e.g. if a vol claims a volunteer multiple times you will end up returning two threads for that thread).

I think what you want is to query volunteer_voter_claims for the latest row with the given volunter_slack_user_id, then join onto that table the needs_attention column from the thread table, which you can then use to filter. I think you want to JOIN on both slack_parent_message_ts and slack_channel_id to make sure you get a one or none match.

src/slack_interaction_handler.ts
```diff
@@ -494,9 +494,91 @@ export async function handleCommandNeedsAttention(
...
+
+    const stats = await DbApiUtil.getThreadsNeedingAttentionByChannel();
+    lines = stats.map(
+      (x) =>
+        `${x.count} in <slack://channel?team=${teamId}&id=${x.channelId}|#${
+          slackChannelNames[x.channelId]
+        }> - oldest ${prettyTimeInterval(x.maxLastUpdateAge)}`
+    );
+    prefix = 'Voters needing attention by channel';
+  } else if (arg && arg[0] == '@') {
```
tomerovadia: Can you put all the arg parsing logic in `command_util`?

src/slack_interaction_handler.ts
```diff
@@ -528,16 +528,18 @@ export async function handleCommandNeedsAttention(
...
-      vstats.map(
-        (x) =>
-          `${x.count} for @${
-            x.volunteerSlackUser
-          } - oldest ${prettyTimeInterval(x.maxLastUpdateAge)}`
-      )
-    );
+    const vstats = await DbApiUtil.getThreadsNeedingAttentionByVolunteer();
+    for (const x of vstats) {
+      const userName = await SlackApiUtil.getSlackUserName(
```
tomerovadia: Can you remove the need for this call by adding slack user name to the DB?

src/slack_api_util.ts
```diff
@@ -355,6 +359,82 @@ export async function updateSlackChannelNamesAndIdsInRedis(
   }
 }
 
+export async function fetchSlackUserNamesAndIds(): Promise<SlackUserNamesAndIds | null> {
```
tomerovadia: Can we write to the DB instead of this?

tomerovadia (commented):
I'm starting to think we're giving the server and Slack app a role that can be fulfilled far better, more simply and with no risk to production critical operations by a basic data analytics tool. The operations for volunteers are great, but for admin commands, I think having admins log into a tool where they can see overall stats makes much more sense than maintaining so much server logic that requires so much complexity for even basic queries. What do you think?

Also, I've been reviewing this commit-by-commit, which I'm realizing has been a huge mistake. I'm a few from the end but really need to shift gears to tonight's release. If you can let me know when you feel it's ready I can take another look at the whole thing altogether.

tomerovadia:
> I'm starting to think we're giving the server and Slack app a role that can be fulfilled far better, more simply and with no risk to production critical operations by a basic data analytics tool. The operations for volunteers are great, but for admin commands, I think having admins log into a tool where they can see overall stats makes much more sense than maintaining so much server logic that requires so much complexity for even basic queries. What do you think?

Perhaps we can move forward with all the features you have here but then use the data analytics tools for any future admin data needs?

sql/20201012.sql
liewegas: removed

src/db_api_util.ts
liewegas: The history_ts is the id of the 'chat history' message that opens up a thread.  It's used only if there are no other messages in the thread, because otherwise when you link to it slack will highlight the message in the channel and not open up a thread view.  This function will prefer to use the newest message in the chat, otherwise the history item as a fallback.

We could always link to the top of the chat (the history), but I think the latest message in the conversation is better!

src/db_api_util.ts
liewegas: fixed in later patches

src/slack_interaction_handler.ts
liewegas: in the empty arg case we show the unclaimed voters for the current channel. new code path is clearer

src/slack_interaction_handler.ts
liewegas: Right now they are posted in the channel.  I think instead it should be privately shown to the user, but then the content has to be returned as a response to the request; will follow up with that later after this is merged

src/db_api_util.ts
liewegas: fixed in latest version

src/async_jobs.ts
liewegas: > > I'm starting to think we're giving the server and Slack app a role that can be fulfilled far better, more simply and with no risk to production critical operations by a basic data analytics tool. The operations for volunteers are great, but for admin commands, I think having admins log into a tool where they can see overall stats makes much more sense than maintaining so much server logic that requires so much complexity for even basic queries. What do you think?
> 
> Perhaps we can move forward with all the features you have here but then use the data analytics tools for any future admin data needs?

Agree... we'll want stats and such but I don't think it makes sense to build all those in.  OTOH, I these commands are the minumum for the volunteers and admins to stay on top of threads and should be integrated directly into the interface.)

external_config/slack.md
liewegas: I'm conflicted... the arguments are only available to admins, not volunteers.

src/slack_interaction_handler.ts
liewegas: clearer in latest code

src/async_jobs.ts
liewegas: fixed

tomerovadia (approved):

PR #125, opened 2020-10-15
Log state_name to messages in Postgres

tomerovadia:
Before regions were a thing, we could derive a voter's U.S. state from their Slack channel name. Even this wasn't ideal, because it would make for difficult analysis if all that was desired was the U.S. state and not demo/non-demo or the pod number. 

But now, with regions, the Slack channel name doesn't even indicate the specific state that a voter selected (e.g. central-0 doesn't let us know the user is in Illinois).

This PR logs the voter's initially selected U.S. state to the messages table in Postgres, for analysis purposes.

I will try to follow up with a migration that will back-fill existing message entries with this data. For existing voters, their subsequent messages should log with the state_name, since stateName is already included as a field in userInfo.

### DIFF
--- a/sql/20201015.sql
+++ b/sql/20201015.sql
@@ -1,0 +1,7 @@
+-- This column represents the U.S. state originally selected by the volunteer.
+-- It is different than the channel name in that:
+-- 1) It doesn't indicate pod number, or demo/non-demo
+-- 2) It is especially important if the channel is a region (e.g. Eastern North)
+-- and the user's U.S. state would be otherwise unknown to the database.
+ALTER TABLE messages
+ADD COLUMN state_name text;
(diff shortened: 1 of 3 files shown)


sql/20201015.sql
```diff
@@ -0,0 +1,7 @@
+-- This column represents the U.S. state originally selected by the volunteer.
+-- It is different than the channel name in that:
+-- 1) It doesn't indicate pod number, or demo/non-demo
+-- 2) It is especially important if the channel is a region (e.g. Eastern North)
+-- and the user's U.S. state would be otherwise unknown to the database.
+ALTER TABLE messages
```
benweissmann, line 6: Do we want an index on this?

benweissmann (approved):

sql/20201015.sql
tomerovadia: Sorry just saw this after I merged. Will follow up with another PR

PR #130, opened 2020-10-16
A few quick thread fixes

liewegas:


src/db_api_util.ts
```diff
@@ -557,14 +557,26 @@ export async function logThreadToDb(
 export async function updateThreadStatusFromMessage(
   databaseMessageEntry: DatabaseMessageEntry
 ): Promise<void> {
+  if (
+    !databaseMessageEntry.slackParentMessageTs ||
+    !databaseMessageEntry.slackChannel
+  ) {
+    // no thread info to update
+    return;
```
benweissmann: When would this case happen? Should we be logging a warning here? It sorta feels like this check should be in whatever's calling `updateThreadStatusFromMessage` so if these fields are unexpectedly missing, we don't silently ignore that.

benweissmann (changes requested):

src/db_api_util.ts
liewegas: Repushed with the check in teh caller.  The errors were happening during the initial welcome and first message of the 'pick your state' exchange, before we have a slack thread.

benweissmann (approved):

PR #134, opened 2020-10-16
Revise welcome messages (shorter, less repetitive)

liewegas:


src/message_constants.ts
```diff
@@ -40,7 +40,7 @@ export function CLARIFY_STATE(): string {
 export function WELCOME_VOTER(): string {
   switch (process.env.CLIENT_ORGANIZATION) {
     case 'VOTE_AMERICA':
-      return 'Welcome to VoteAmerica! We are excited to help you vote and can connect you to a trained volunteer to help answer your questions.\n\nReply HELPLINE to be connected to a volunteer. Reply STOP to unsubscribe from VoteAmerica texts. (Msg & data rates may apply).';
+      return "Welcome to VoteAmerica! We're excited to help you vote.\n\nReply HELPLINE to speak to a trained volunteer. Reply STOP to unsubscribe. (Msg&data rates may apply)";
```
tomerovadia: This is great. I'm a bit concerned that a voter might think "speak" implies a phone call. How about:

Welcome to VoteAmerica! We're excited to help you vote.\n\nReply HELPLINE to connect with a volunteer. Reply STOP to unsubscribe. (Msg & data rates may apply).

liewegas: updated. also modified the welcome + state message to fit in 160 chars

src/message_constants.ts
```diff
@@ -40,7 +40,7 @@ export function CLARIFY_STATE(): string {
 export function WELCOME_VOTER(): string {
   switch (process.env.CLIENT_ORGANIZATION) {
     case 'VOTE_AMERICA':
-      return 'Welcome to VoteAmerica! We are excited to help you vote and can connect you to a trained volunteer to help answer your questions.\n\nReply HELPLINE to be connected to a volunteer. Reply STOP to unsubscribe from VoteAmerica texts. (Msg & data rates may apply).';
+      return "Welcome to VoteAmerica! We're excited to help you vote.\n\nReply HELPLINE to connect with a trained volunteer, or STOP to unsubscribe. (Msg&data rates may apply)";
```
tomerovadia, line 43: By my count this is 161 characters

src/message_constants.ts
```diff
@@ -72,7 +72,7 @@ export function CLARIFY_HELPLINE_REQUEST(): string {
 
 // Only used by VOTE_AMERICA.
 export function WELCOME_AND_STATE_QUESTION(): string {
-  return 'Welcome to VoteAmerica! We are excited to help you vote and can connect you to a trained volunteer to help answer your questions.\n\nTo match you with the most knowledgeable volunteer, in which U.S. state are you looking to vote? (Msg & data rates may apply)\n\nReply STOP to unsubscribe from VoteAmerica texts.';
+  return 'Welcome to the VoteAmerica Helpline! We are excited to help. Msg&data rates may apply. Reply STOP to unsubscribe.\n\nIn which U.S. state are you looking to vote?';
```
tomerovadia, line 75: This also comes out to 161 characters by my count

tomerovadia, line 75: I think this is a great idea and this copy looks great to me (if we can squeeze into 160 chars). But can you run the final drafts by Snook and Brenna to make sure?

liewegas: I think you're off by the escape characters? Both are just under 160, and (re)confirmed from twilio logs that they're fitting in a single segment

tomerovadia: Ya you're right, thanks!

tomerovadia (changes requested):
Can you change the PR name so that it includes the removal of the double-great? That way we can more easily see what changes happened with each PR / find changes.

tomerovadia (approved):

PR #135, opened 2020-10-16
/unclaimed: ignore spam|refused

liewegas:


src/db_api_util.ts
```diff
@@ -700,38 +700,52 @@ export async function getUnclaimedVoters(
   const client = await pool.connect();
   try {
     let result = null;
+    const withPreamble = `WITH all_status AS (
```
benweissmann, line 703: I'm a little worries about the performance of this query -- what does `EXPLAIN ANALYZE` say about how this is going to perform vs. the query without this preamble?

liewegas: It's definitely slower.  It goes from ~.5ms to ~5ms:
```
 Sort  (cost=87.36..87.42 rows=24 width=78) (actual time=0.523..0.526 rows=63 loops=1)
   Sort Key: t.updated_at
   Sort Method: quicksort  Memory: 33kB
   ->  Hash Anti Join  (cost=44.53..86.81 rows=24 width=78) (actual time=0.327..0.501 rows=63 loops=1)
         Hash Cond: (t.user_id = c.user_id)
         ->  Index Scan using threads_needs_attention_idx on threads t  (cost=0.28..38.33 rows=179 width=70) (actual time=0.021..0.157 rows=179 loops=1)
               Index Cond: (needs_attention = true)
               Filter: needs_attention
         ->  Hash  (cost=33.00..33.00 rows=900 width=33) (actual time=0.290..0.290 rows=924 loops=1)
               Buckets: 1024  Batches: 1  Memory Usage: 67kB
               ->  Seq Scan on volunteer_voter_claims c  (cost=0.00..33.00 rows=900 width=33) (actual time=0.004..0.152 rows=924 loops=1)
 Planning Time: 0.338 ms
 Execution Time: 0.559 ms
```
to
```
 Sort  (cost=226.25..226.26 rows=1 width=78) (actual time=4.249..4.249 rows=0 loops=1)
   Sort Key: t.updated_at
   Sort Method: quicksort  Memory: 25kB
   CTE all_status
     ->  WindowAgg  (cost=116.50..146.32 rows=1491 width=53) (actual time=2.288..3.104 rows=1465 loops=1)
           ->  Sort  (cost=116.50..120.23 rows=1491 width=45) (actual time=2.283..2.353 rows=1465 loops=1)
                 Sort Key: voter_status_updates.user_id, voter_status_updates.created_at DESC
                 Sort Method: quicksort  Memory: 255kB
                 ->  Seq Scan on voter_status_updates  (cost=0.00..37.91 rows=1491 width=45) (actual time=0.005..0.215 rows=1465 loops=1)
   CTE latest_status
     ->  CTE Scan on all_status  (cost=0.00..33.55 rows=7 width=36) (actual time=2.290..3.548 rows=981 loops=1)
           Filter: (rn = 1)
           Rows Removed by Filter: 484
   ->  Nested Loop Anti Join  (cost=0.55..46.38 rows=1 width=78) (actual time=4.246..4.246 rows=0 loops=1)
         ->  Nested Loop  (cost=0.28..39.00 rows=1 width=70) (actual time=2.820..4.223 rows=2 loops=1)
               Join Filter: (t.user_id = s.user_id)
               Rows Removed by Join Filter: 1844
               ->  Index Scan using threads_needs_attention_idx on threads t  (cost=0.28..38.78 rows=1 width=70) (actual time=0.150..0.153 rows=2 loops=1)
                     Index Cond: (needs_attention = true)
                     Filter: (needs_attention AND (slack_channel_id = 'G01BQDV3G4W'::text))
                     Rows Removed by Filter: 176
               ->  CTE Scan on latest_status s  (cost=0.00..0.16 rows=5 width=32) (actual time=1.147..1.971 rows=923 loops=2)
                     Filter: (voter_status <> ALL ('{REFUSED,SPAM}'::voter_status[]))
                     Rows Removed by Filter: 58
         ->  Index Only Scan using idx_volunteer_voter_claims_user_id on volunteer_voter_claims c  (cost=0.28..8.29 rows=1 width=33) (actual time=0.010..0.010 rows=1 loops=2)
               Index Cond: (user_id = t.user_id)
               Heap Fetches: 2
 Planning Time: 0.497 ms
 Execution Time: 4.316 ms
```
I think it's okay for now at least.. everything is indexed.  If we need to, though, we could mirror the spam/refused state in the threads table.

tomerovadia (approved):

src/db_api_util.ts
benweissmann: Adding a compound index to eliminate the seq scan on voter_status_updates brings that down to about 2ms... probably worth adding?

`CREATE INDEX voter_status_updates_user_id_created_at ON voter_status_updates(user_id, created_at DESC);`

PR #146, opened 2020-10-18
Have Redis provide StateRegionConfig

tomerovadia:
Right now, every time we want to update the StateRegionConfig we have to go through the normal release cycle, which takes time, testing and often needs syncing with whatever other code changes are happening.

This PR instead has Redis provide the StateRegionConfig, so that any changes are easily made much more quickly via Redis, not requiring a release.


src/state_region_config.ts
```diff
@@ -1,67 +1,34 @@
...
+  );
+  if (stateRegionConfig) return stateRegionConfig;
+  return {} as StateRegionConfig;
+}
+
+export async function regionsList(
+  redisClient: PromisifiedRedisClient
+): Promise<string[]> {
+  return process.env.CLIENT_ORGANIZATION === 'VOTE_AMERICA'
+    ? uniq(Object.values(fetchStateRegionConfig(redisClient)))
```
fongandrew: We can probably remove the client org check if it's in Redis anyway?

fongandrew (approved):

src/state_region_config.ts
tomerovadia: The other client uses a separate Redis. Do you know if I would need to update their Redis as well for this to work properly?

fongandrew: It should? I assume it's an empty hash on the other client, so this function will just return an empty list.

tomerovadia: Good point. I'll test that out and follow up.

tomerovadia: It works! Thanks for suggesting that. I pushed a commit with the change, as well as another small but significant change to how fallbacks work if Redis is lacking needed information. Please take a look

src/state_region_config.ts
```diff
@@ -21,9 +21,10 @@ export async function fetchStateRegionConfig(
 export async function regionsList(
   redisClient: PromisifiedRedisClient
 ): Promise<string[]> {
-  return process.env.CLIENT_ORGANIZATION === 'VOTE_AMERICA'
-    ? uniq(Object.values(fetchStateRegionConfig(redisClient)))
-    : [];
+  const stateRegionConfig = await fetchStateRegionConfig(redisClient);
+  console.log("TOMERTOMERTOMERTOMERTOMERTOMEROMTERTOMERTOMERTOMER");
```
fongandrew: Stray console log

fongandrew:
Just the stray log but everything else looks fine to me.

src/state_region_config.ts
tomerovadia: Thanks! Removed

fongandrew (approved):

PR #156, opened 2020-10-20
Handle VOTED messages

liewegas:
- Identify `VOTED`, `I voted`, `I've already voted`, etc.
- Log voter status
- Adjust blocks on new threads for those who previously texted VOTED

### DIFF
--- a/sql/20201021.sql
+++ b/sql/20201021.sql
@@ -1,2 +1,2 @@
--- Add VOTED_SMS to enum values for voter_status
+-- Add ALREADY_VOTED to enum values for voter_status
 ALTER TYPE voter_status ADD VALUE 'ALREADY_VOTED';
--- a/sql/20201021.sql
+++ b/sql/20201021.sql
@@ -1,0 +1,2 @@
+-- Add VOTED_SMS to enum values for voter_status
+ALTER TYPE voter_status ADD VALUE 'ALREADY_VOTED';
--- a/src/db_api_util.ts
+++ b/src/db_api_util.ts
@@ -1032,6 +1032,7 @@ export async function logInitialVoterStatusToDb(
   const client = await pool.connect();
   try {
     if (process.env.CLIENT_ORGANIZATION === 'VOTE_AMERICA') {
+      // Only insert the UNKNOWN status if there is no existing (ALREADY_VOTED) status
       await client.query(
         `INSERT INTO voter_status_updates (user_id, user_phone_number, voter_status, is_demo)
         SELECT $1, $2, 'UNKNOWN', $3
--- a/src/db_api_util.ts
+++ b/src/db_api_util.ts
@@ -7,7 +7,6 @@ import {
   HistoricalMessage,
   SlackThreadInfo,
 } from './types';
-import { phoneNumbersAreDemo } from './load_balancer';
 
 const pool = new Pool({
   connectionString: process.env.DATABASE_URL,
@@ -1026,12 +1025,12 @@ export async function logVoterStatusToDb(
 export async function logInitialVoterStatusToDb(
   userId: string,
   userPhoneNumber: string,
-  twilioPhoneNumber: string
+  twilioPhoneNumber: string,
+  isDemo: boolean
 ): Promise<void> {
   logger.info(`ENTERING DBAPIUTIL.logInitialVoterStatusToDb`);
   const client = await pool.connect();
   try {
-    const isDemo = phoneNumbersAreDemo(twilioPhoneNumber, userPhoneNumber);
     await client.query(
       `INSERT INTO voter_status_updates (user_id, user_phone_number, voter_status, is_demo)
       SELECT $1, $2, 'UNKNOWN', $3
--- a/src/db_api_util.ts
+++ b/src/db_api_util.ts
@@ -7,6 +7,8 @@ import {
   HistoricalMessage,
   SlackThreadInfo,
 } from './types';
+import { phoneNumbersAreDemo } from './load_balancer';
+
 const pool = new Pool({
   connectionString: process.env.DATABASE_URL,
   max: Number(process.env.CONNECTION_POOL_MAX || 20),
@@ -1029,8 +1031,7 @@ export async function logInitialVoterStatusToDb(
   logger.info(`ENTERING DBAPIUTIL.logInitialVoterStatusToDb`);
   const client = await pool.connect();
   try {
-    const isDemo =
-      twilioPhoneNumber === process.env.DEMO_PHONE_NUMBER ? true : null;
+    const isDemo = phoneNumbersAreDemo(twilioPhoneNumber, userPhoneNumber);
     await client.query(
       `INSERT INTO voter_status_updates (user_id, user_phone_number, voter_status, is_demo)
       SELECT $1, $2, 'UNKNOWN', $3
--- a/src/db_api_util.ts
+++ b/src/db_api_util.ts
@@ -1036,9 +1036,9 @@ export async function logInitialVoterStatusToDb(
       SELECT $1, $2, 'UNKNOWN', $3
       WHERE NOT EXISTS (
         SELECT null FROM voter_status_updates
-        WHERE user_id = $4 AND user_phone_number = $5
+        WHERE user_id = $1 AND user_phone_number = $2 AND is_demo = $3
       )`,
-      [userId, userPhoneNumber, isDemo, userId, userPhoneNumber]
+      [userId, userPhoneNumber, isDemo]
     );
     logger.info(
       `DBAPIUTIL.logInitialVoterStatusToDb: Successfully inserted initial voter status into PostgreSQL database.`
--- a/src/db_api_util.ts
+++ b/src/db_api_util.ts
@@ -1021,6 +1021,33 @@ export async function logVoterStatusToDb(
   }
 }
 
+export async function logInitialVoterStatusToDb(
+  userId: string,
+  userPhoneNumber: string,
+  twilioPhoneNumber: string
+): Promise<void> {
+  logger.info(`ENTERING DBAPIUTIL.logInitialVoterStatusToDb`);
+  const client = await pool.connect();
+  try {
+    const isDemo =
+      twilioPhoneNumber === process.env.DEMO_PHONE_NUMBER ? true : null;
+    await client.query(
+      `INSERT INTO voter_status_updates (user_id, user_phone_number, voter_status, is_demo)
+      SELECT $1, $2, 'UNKNOWN', $3
+      WHERE NOT EXISTS (
+        SELECT null FROM voter_status_updates
+        WHERE user_id = $4 AND user_phone_number = $5
+      )`,
+      [userId, userPhoneNumber, isDemo, userId, userPhoneNumber]
+    );
+    logger.info(
+      `DBAPIUTIL.logInitialVoterStatusToDb: Successfully inserted initial voter status into PostgreSQL database.`
+    );
+  } finally {
+    client.release();
+  }
+}
+
 const LAST_VOTER_STATUS_SQL_SCRIPT = `SELECT voter_status
                                         FROM voter_status_updates
                                         WHERE user_id = $1
--- a/src/db_api_util.ts
+++ b/src/db_api_util.ts
@@ -1031,15 +1031,23 @@ export async function logInitialVoterStatusToDb(
   logger.info(`ENTERING DBAPIUTIL.logInitialVoterStatusToDb`);
   const client = await pool.connect();
   try {
-    await client.query(
-      `INSERT INTO voter_status_updates (user_id, user_phone_number, voter_status, is_demo)
-      SELECT $1, $2, 'UNKNOWN', $3
-      WHERE NOT EXISTS (
-        SELECT null FROM voter_status_updates
-        WHERE user_id = $1 AND user_phone_number = $2 AND is_demo = $3
-      )`,
-      [userId, userPhoneNumber, isDemo]
-    );
+    if (process.env.CLIENT_ORGANIZATION === 'VOTE_AMERICA') {
+      await client.query(
+        `INSERT INTO voter_status_updates (user_id, user_phone_number, voter_status, is_demo)
+        SELECT $1, $2, 'UNKNOWN', $3
+        WHERE NOT EXISTS (
+          SELECT null FROM voter_status_updates
+          WHERE user_id = $1 AND user_phone_number = $2 AND is_demo = $3
+        )`,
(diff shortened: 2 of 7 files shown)


sql/20201021.sql
```diff
@@ -0,0 +1,2 @@
+-- Add VOTED_SMS to enum values for voter_status
```
tomerovadia: VOTED_SMS or VOTED_TEXT?

What are your thoughts on naming it something that distinguishes a bit more from other voted statuses (e.g. `ALREADY_VOTED`)?

src/message_constants.ts
```diff
@@ -40,7 +40,7 @@ export function CLARIFY_STATE(): string {
 export function WELCOME_VOTER(): string {
   switch (process.env.CLIENT_ORGANIZATION) {
     case 'VOTE_AMERICA':
-      return "Welcome to VoteAmerica! We're excited to help you vote.\n\nReply HELPLINE to connect with a trained volunteer, or STOP to unsubscribe. (Msg&data rates may apply)";
+      return "Welcome to VoteAmerica! Msg&data rates may apply.\n\nReply HELPLINE to connect with a trained volunteer, VOTED if you've already voted, STOP to unsubscribe.";
```
tomerovadia, line 43: Looks good to me. Can we run this by Brenna/Snook?

src/message_constants.ts
```diff
@@ -72,7 +72,11 @@ export function CLARIFY_HELPLINE_REQUEST(): string {
 
 // Only used by VOTE_AMERICA.
 export function WELCOME_AND_STATE_QUESTION(): string {
-  return 'Welcome to the VoteAmerica Helpline! We are excited to help. Msg&data rates may apply. Reply STOP to unsubscribe.\n\nIn which U.S. state are you looking to vote?';
+  return "Welcome to the VoteAmerica Helpline! Msg&data rates may apply.\n\nReply VOTED if you've alread voted, STOP to unsubscribe.\n\nIn which U.S. state will you vote?";
```
tomerovadia: Typo (alread -> already). Also, can you run this by Brenna/Snook?

Can we remove `Helpline` and put back `are you looking to vote`?

src/router.ts
```diff
@@ -20,6 +20,16 @@ import * as Sentry from '@sentry/node';
 const MINS_BEFORE_WELCOME_BACK_MESSAGE = 60 * 24;
 export const NUM_STATE_SELECTION_ATTEMPTS_LIMIT = 2;
 
+const VOTED_KEYWORDS = [
```
tomerovadia: Can you add these to the keyword_parser?

src/router.ts
```diff
@@ -31,6 +41,10 @@ type AdminCommandParams = {
   previousSlackChannelName: string;
 };
 
+export function isVotedMessage(message: string): boolean {
```
tomerovadia: Same here -- can this go in the keyword_parser?

src/router.ts
```diff
@@ -86,8 +100,31 @@ export async function welcomePotentialVoter(
     entryPoint,
   });
 
+  let message = MessageConstants.WELCOME_VOTER();
```
tomerovadia: can you rename this `messageToVoter `, to clearly differentiate from the user message and for consistency with other parts of the file?

src/router.ts
```diff
@@ -86,8 +100,31 @@ export async function welcomePotentialVoter(
...
+  if (
+    process.env.CLIENT_ORGANIZATION === 'VOTE_AMERICA' &&
+    isVotedMessage(inboundDbMessageEntry.message || '')
```
tomerovadia: Let's use `userOptions.userMessage` instead of `inboundDbMessageEntry.message` and I believe you can omit the `|| ''`

src/router.ts
```diff
@@ -86,8 +100,31 @@ export async function welcomePotentialVoter(
...
+  ) {
+    message = MessageConstants.VOTED_WELCOME_RESPONSE();
+    await DbApiUtil.logVoterStatusToDb({
+      userId: userInfo.userId,
+      userPhoneNumber: userInfo.userPhoneNumber,
+      twilioPhoneNumber: twilioPhoneNumber,
+      isDemo: LoadBalancer.phoneNumbersAreDemo(
```
tomerovadia: you should use `userInfo.isDemo` for isDemo

src/router.ts
```diff
@@ -86,8 +100,31 @@ export async function welcomePotentialVoter(
     entryPoint,
   });
 
+  let message = MessageConstants.WELCOME_VOTER();
+  if (
+    process.env.CLIENT_ORGANIZATION === 'VOTE_AMERICA' &&
```
tomerovadia: You can omit the CLIENT_ORGANIZATION check since it's already done here: https://github.com/tomerovadia/VoterHelpLine/blob/master/src/app.ts#L431

src/router.ts
```diff
@@ -144,7 +181,20 @@ const introduceNewVoterToSlackChannel = async (
   // In Slack, create entry channel message, followed by voter's message and intro text.
   const operatorMessage = `*User ID:* ${userInfo.userId}\n*Connected via:* ${twilioPhoneNumber} (${entryPoint})`;
 
-  const slackBlocks = SlackBlockUtil.getVoterStatusBlocks(operatorMessage);
+  let slackBlocks = SlackBlockUtil.getVoterStatusBlocks(operatorMessage);
```
tomerovadia: I thought we're not handling VOTED messages once they're in the Slack workspace?

src/db_api_util.ts
```diff
@@ -1021,6 +1021,33 @@ export async function logVoterStatusToDb(
...
+  twilioPhoneNumber: string
+): Promise<void> {
+  logger.info(`ENTERING DBAPIUTIL.logInitialVoterStatusToDb`);
+  const client = await pool.connect();
+  try {
+    const isDemo =
+      twilioPhoneNumber === process.env.DEMO_PHONE_NUMBER ? true : null;
+    await client.query(
+      `INSERT INTO voter_status_updates (user_id, user_phone_number, voter_status, is_demo)
+      SELECT $1, $2, 'UNKNOWN', $3
```
tomerovadia: Should there be a `SELECT` clause in your `INSERT INTO` clause? Are you trying to populate voter_status_update based on another table? Should there be a `FROM` clause?

src/db_api_util.ts
```diff
@@ -1021,6 +1021,33 @@ export async function logVoterStatusToDb(
...
+export async function logInitialVoterStatusToDb(
+  userId: string,
+  userPhoneNumber: string,
+  twilioPhoneNumber: string
+): Promise<void> {
+  logger.info(`ENTERING DBAPIUTIL.logInitialVoterStatusToDb`);
+  const client = await pool.connect();
+  try {
+    const isDemo =
+      twilioPhoneNumber === process.env.DEMO_PHONE_NUMBER ? true : null;
```
tomerovadia: all isDemo determinations should go through the LoadBalancer: https://github.com/tomerovadia/VoterHelpLine/blob/master/src/load_balancer.ts#L12

src/router.ts
```diff
@@ -325,18 +375,11 @@ export async function handleNewVoter(
...
-    twilioPhoneNumber,
-    voterStatus: 'UNKNOWN',
-    originatingSlackUserName: null,
-    originatingSlackUserId: null,
-    slackChannelName: null,
-    slackChannelId: null,
-    slackParentMessageTs: null,
-    isDemo: userInfo.isDemo || null,
-  });
+  await DbApiUtil.logInitialVoterStatusToDb(
```
tomerovadia, line 357: Is this a refactor, or related to the rest of the PR? handleNewVoter handles inserting a voter into the Slack workspace, but I thought we were only changing logic that runs for voters before they enter

src/router.ts
tomerovadia: can you strip away punctuation before doing the check, and update the VOTED_KEYWORDS accordingly? there's an example in the keyword_parser. That way if someone says `I've voted already!!!` we can catch that.

tomerovadia (commented):
Can you update the description? Added a bunch of comments that hopefully make sense. We can discuss via Slack or Zoom if needed.

src/router.ts
liewegas: We're not, but we need to set the initial blocks to reflect that they previously said they'd voted when they enter the helpline.

src/db_api_util.ts
liewegas: postgres doesn't have a ``REPLACE INTO`` ala mysql; this is the way to do a conditional insert, so that we only insert UNKNOWN if the ALREADY_VOTED isn't already present.

src/router.ts
liewegas: Nope, it's because we need the conditional UNKNOWN status insert

sql/20201021.sql
tomerovadia: I think you've got an outdated comment (`VOTED_SMS` instead of `ALREADY_VOTED`)

src/db_api_util.ts
tomerovadia: sounds good. can you add that as a comment? and am I understanding that this won't insert if no existing `voter_status_update` is found for the user? if so, can you add that in the comment as well?

src/router.ts
tomerovadia: I see okay that makes sense. Can you flesh out the comment to make that clearer? And am I understanding correctly that voters will only enter the helpline and have this block changed if they texted VOTED (or something similar) and then later texted HELPLINE for some reason (e.g. something went wrong, or they have a question for a friend, etc)? If so, can you add something to that effect to the comment too?

src/router.ts
```diff
@@ -144,7 +162,20 @@ const introduceNewVoterToSlackChannel = async (
...
   const operatorMessage = `*User ID:* ${userInfo.userId}\n*Connected via:* ${twilioPhoneNumber} (${entryPoint})`;
 
-  const slackBlocks = SlackBlockUtil.getVoterStatusBlocks(operatorMessage);
+  let slackBlocks = SlackBlockUtil.getVoterStatusBlocks(operatorMessage);
+
+  // already voted?
+  const status = await DbApiUtil.getLatestVoterStatus(userInfo.userId);
+  if (status === 'ALREADY_VOTED') {
+    const votedBlocks = SlackBlockUtil.makeClosedVoterPanelBlocks(
+      'Voter marked *VOTED* because they texted `VOTED` (or similar)',
```
tomerovadia: I worry we might unintentionally ignore these messages because they have closed voter panel blocks, even though if I'm understanding correctly these people will have texted HELPLINE after texting VOTED. The intention of the closed voter blocks is to communicate to a volunteer that this voter is done and doesn't need any more follow-up, which IIUC isn't the case in these situations. 

Instead of closing the voter blocks, can we simply change the dropdown to Already Voted? After a volunteer sees what's going on and helps out and if the voter is all good to go, they can click the VOTED button and the blocks can then close (and the status would change to VOTED instead of ALREADY VOTED, differentiating from voters we didn't help who texted VOTED). 

You'd have to update this as well: https://github.com/tomerovadia/VoterHelpLine/blob/8c8a2dec38f5647c6a040b98623d4a9709512b3c/src/slack_block_util.ts#L65

The logic for a dropdown change is here: https://github.com/tomerovadia/VoterHelpLine/blob/master/src/slack_interaction_handler.ts#L252

src/db_api_util.ts
```diff
@@ -1021,6 +1023,32 @@ export async function logVoterStatusToDb(
...
 
+export async function logInitialVoterStatusToDb(
+  userId: string,
+  userPhoneNumber: string,
+  twilioPhoneNumber: string
+): Promise<void> {
+  logger.info(`ENTERING DBAPIUTIL.logInitialVoterStatusToDb`);
+  const client = await pool.connect();
+  try {
+    const isDemo = phoneNumbersAreDemo(twilioPhoneNumber, userPhoneNumber);
```
tomerovadia: can you pass `userInfo.isDemo` instead of recalculating here?

src/router.ts
tomerovadia: Gotcha sounds good. Thanks for thinking of this. Can we do this conditionally on CLIENT_ORAGANIZATION=VOTE_AMERICA? That way the other organization that isn't using the ALREADY_VOTED status doesn't need to do the `voter_status_updates` DB lookup.

src/router.ts
liewegas: Yes! That'd be much better.

liewegas: Okay, this is updated.  The weirdness now is that the selector has both an `Already voted` option and there is a `Voted` button to the right.  Is this a useful distinction?  I think we should consider having just a single `Voted` state (in the dropdown) and get rid of the button.  Even if someone votes via the helpline they may come back with more questions, just like if they texted in.  Also, the `Voted` button has the somewhat annoying modal you have to wait for and click to confirm.

sql/20201021.sql
tomerovadia: It still says `VOTED_SMS` for me

src/router.ts
```diff
@@ -146,6 +164,15 @@ const introduceNewVoterToSlackChannel = async (
 
   const slackBlocks = SlackBlockUtil.getVoterStatusBlocks(operatorMessage);
 
+  // already voted?
+  const status = await DbApiUtil.getLatestVoterStatus(userInfo.userId);
+  if (status === 'ALREADY_VOTED') {
+    slackBlocks[2].elements[0].initial_option =
```
tomerovadia: IIUC this is assuming that the dropdown is the third block and first element, right? can you use can you use populateDropdownNewInitialValue https://github.com/tomerovadia/VoterHelpLine/blob/09f87a94b712094a9f9631e908e8d292492ba873/src/slack_block_util.ts#L459 so we're resilient to changes in the block structure down the line?

You can see an example here: https://github.com/tomerovadia/VoterHelpLine/blob/master/src/slack_interaction_handler.ts#L257

Also, you need to call replaceSlackMessageBlocks or the change does not persist https://github.com/tomerovadia/VoterHelpLine/blob/master/src/slack_interaction_handler.ts#L271

liewegas: switched to the helper.

no need to replace blocks since this is when we're creating them for the first time

src/db_api_util.ts
```diff
@@ -1032,6 +1032,7 @@ export async function logInitialVoterStatusToDb(
   const client = await pool.connect();
   try {
     if (process.env.CLIENT_ORGANIZATION === 'VOTE_AMERICA') {
+      // Only insert the UKNONWN status if there is no existing (ALREADY_VOTED) status
```
tomerovadia: typo (UKNONWN)

tomerovadia (approved):

PR #160, opened 2020-10-21
Make automated messages use custom icon/name; avoid full md5's

liewegas:


src/router.ts
```diff
@@ -969,6 +969,7 @@ export async function handleDisclaimer(
   const slackLobbyMessageParams = {
     parentMessageTs: userInfo[userInfo.activeChannelId],
     channel: userInfo.activeChannelId,
+    isAutomatedMessage: true,
```
tomerovadia: This isn't for an automated message necessarily, right? Looks like it's used below to relay the voter's message

liewegas: fixed!

tomerovadia (approved):
Looks great!

PR #163, opened 2020-10-22
Outbound MMS/attachment support

liewegas:
Based on @benweissmann 's initial patch.  Some hackery to construct a usable URL (slack API doesn't make it easy to get one we can pass to twilio).

There's an extra slack configuration step: you need to create a `OAuth Access Token` (in addition to the `Bot User OAuth Access Token`) and put it in .env.  Also, a migration to add a new column.


src/db_api_util.ts
```diff
@@ -153,6 +155,12 @@ export async function logMessageToDb(
         databaseMessageEntry.entryPoint,
         databaseMessageEntry.archived,
         databaseMessageEntry.stateName,
+
+        // We can't just pass the array as-is because this is a JSON column,
+        // an node-pg serializes arrays as PG arrays rather than JSON arrays.
```
tomerovadia: nit: an -> and?

src/router.ts
```diff
@@ -1140,6 +1141,31 @@ export async function handleSlackVoterThreadMessage(
...
+  outboundDbMessageEntry.slackFiles = reqBody.event.files;
+
+  // check attachments
+  if (outboundDbMessageEntry.slackFiles) {
+    const errors = MessageParser.validateSlackAttachments(
+      outboundDbMessageEntry.slackFiles
+    );
+    if (errors.length) {
+      await SlackApiUtil.sendMessage(
+        'Sorry, there was a problem with one or more of your attachments:\n' +
```
tomerovadia, line 1174: Is there a way to test that this works correctly (and if so, can you add a screenshot to the pull request description)?

src/slack_api_util.ts
```diff
@@ -71,6 +72,25 @@ export async function getThreadPermalink(
...
+      null,
+      {
+        params: {
+          file: file.id,
+          token: process.env.SLACK_USER_ACCESS_TOKEN,
+        },
+      }
+    );
+    if (!response.data.ok) {
+      logger.error(`SLACKAPIUTIL.makeFilesPublic failed with ${response.data}`);
```
tomerovadia: can you stringify `response.data`? I believe it otherwise appears as `[Object object]`

src/twilio_util.ts
```diff
@@ -57,5 +57,5 @@ export function formatAttachments(reqBody: TwilioRequestBody): string {
     .map((url, i) => `<${url}|Attachment ${i + 1}>`)
     .join(' ');
 
-  return `${reqBody.Body} [This message included attachments: ${mediaURLsFormatted}]`;
+  return `${reqBody.Body}\n[${mediaURLsFormatted}]`;
```
tomerovadia: I think the [This message included attachments:] bit helps add clarity as to where the link came from (ie to differentiate from the user sending a link themselves).

src/twilio_api_util.ts
```diff
@@ -55,11 +56,19 @@ export async function sendMessage(
   }
 
   try {
+    let media = [] as string[];
+    if (databaseMessageEntry.slackFiles) {
+      media = MessageParser.getSlackAttachments(
+        databaseMessageEntry.slackFiles
+      );
+      logger.info(`Including attachments ${JSON.stringify(media)}`);
```
tomerovadia: Can you add `TWILIOAPIUTIL.sendMessage: ` to the start of this log message?

src/slack_api_util.ts
```diff
@@ -71,6 +72,25 @@ export async function getThreadPermalink(
...
+    }
+    logger.info('SLACKAPIUTIL: successfully make ${file.id} public');
```
tomerovadia: Can you make the start of this log statement `SLACKAPIUTIL.makeFilesPublic:`

src/message_parser.ts
```diff
@@ -4,6 +4,50 @@ const emoji = new EmojiConvertor();
...
+
+// From: https://www.twilio.com/docs/sms/accepted-mime-types#supported-mime-types
+const SUPPORTED_MMS_MIME_TYPES = ['image/jpeg', 'image/gif', 'image/png'];
+
+// Documented size limit is 5MB. To give us a bit of buffer, we assume they
+// calculate MB with 1000 instead of 1024, and we limit to 90% of this
+// hard cap.
+const MMS_MAX_SIZE = 1000 * 1000 * 5 * 0.9;
+
 export default {
```
tomerovadia: @benweissmann might know, but should we remove the `export default` now that we have multiple exports?

src/db_api_util.ts
```diff
@@ -123,7 +125,7 @@ export async function logMessageToDb(
...
+      'INSERT INTO messages (message, direction, automated, successfully_sent, from_phone_number, user_id, to_phone_number, originating_slack_user_id, slack_channel, slack_parent_message_ts, twilio_message_sid, slack_message_ts, slack_error, twilio_error, twilio_send_timestamp, twilio_receive_timestamp, slack_send_timestamp, slack_receive_timestamp, confirmed_disclaimer, is_demo, last_voter_message_secs_from_epoch, unprocessed_message, slack_retry_num, slack_retry_reason, originating_slack_user_name, entry_point, archived, state_name, slack_files) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20, $21, $22, $23, $24, $25, $26, $27, $28, $29);',
```
tomerovadia: nit: name `slack_attachments`?

src/router.ts
```diff
@@ -1140,6 +1141,31 @@ export async function handleSlackVoterThreadMessage(
     slackRetryNum: retryCount,
     slackRetryReason: retryReason,
   });
+  outboundDbMessageEntry.slackFiles = reqBody.event.files;
+
+  // check attachments
+  if (outboundDbMessageEntry.slackFiles) {
```
tomerovadia: The `outboundDbMessageEntry` shouldn't be responsible for carrying data between functions except for the purpose of logging to the database. Can you pass `slackFiles` via the `options` object instead?

tomerovadia (commented):
This mostly looks good to me. Added some comments. Also adding @fongandrew as a reviewer if he doesn't mind, since this PR is heavy with Slack specifics. Also, did you forget to include a migration?

fongandrew:
Slack bits look good to me. We talked about this in thread earlier.

src/router.ts
liewegas: image

src/message_parser.ts
liewegas: seems more consistent w/ the rest of the codebase.. will change!

PR #165, opened 2020-10-22
get originating slack username for interactivity from the payload

liewegas:
Thanks @eliblock for the suggestion!

### DIFF
--- a/src/async_jobs.ts
+++ b/src/async_jobs.ts
@@ -82,14 +82,7 @@ async function slackInteractivityHandler(
   payload: SlackInteractionEventPayload,
   interactivityMetadata: InteractivityHandlerMetadata
 ) {
-  const originatingSlackUserName = await SlackApiUtil.fetchSlackUserName(
-    payload.user.id
-  );
-  if (!originatingSlackUserName) {
-    throw new Error(
-      `Could not get slack user name for slack user ${payload.user.id}`
-    );
-  }
+  const originatingSlackUserName = payload.user.username;
 
   // Global shortcut
   if (payload.type === 'shortcut') {
--- a/src/slack_interaction_handler.ts
+++ b/src/slack_interaction_handler.ts
@@ -54,6 +54,7 @@ export type SlackInteractionEventPayload = {
   actions: SlackBlockUtil.SlackAction[];
   user: {
     id: string;
+    username: string;
   };
   message: {
     ts: string;


src/async_jobs.ts
```diff
@@ -82,14 +82,7 @@ async function slackInteractivityHandler(
...
 ) {
-  const originatingSlackUserName = await SlackApiUtil.fetchSlackUserName(
-    payload.user.id
-  );
-  if (!originatingSlackUserName) {
-    throw new Error(
-      `Could not get slack user name for slack user ${payload.user.id}`
-    );
-  }
+  const originatingSlackUserName = payload.user.username;
```
tomerovadia, line 85: This is a different username. `fetchSlackUserName` returns a human recognizable username (to allow volunteers to identify and DM one another if needed). `payload.user.username` returns something I believe is affiliated with account information.

For example, here is a payload from an interaction by me (even though my normal name is in my profile for both `Full Name` and `Display Name`:

`
{
  type: 'message_action',
  action_ts: '1603421998.331787',
  user: {
    id: 'U017R2YDDAN',
    username: 'tomertomertomer',
    team_id: 'T017R90PJKV',
    **name: 'tomertomertomer'**
  },
  channel: { id: 'C018ETNSMKJ', name: 'privategroup' },
  callback_id: 'reset_demo',
  trigger_id: '1433427322615.1263306800675.873b92d0b0a5ab775a4b878f3243ac3c',
  response_url: 'https://hooks.slack.com/app/T017R90PJKV/1448167480418/iKKaQMJvertBUciyLwyMgF0s',
  message_ts: '1603421882.000700',
  message: {
    type: 'message',
    user: 'U017PMHETGD',
    ts: '1603421882.000700',
    team: 'T017R90PJKV',
    bot_profile: {
      deleted: false,
      name: 'Voter Help Line Staging',
      updated: 1595733113,
      icons: [Object],
    },
    blocks: [ [Object], [Object], [Object] ],
    thread_ts: '1603421882.000700',
    reply_count: 1,
    reply_users_count: 1,
    latest_reply: '1603421882.000800',
    reply_users: [ 'U017PMHETGD' ],
    subscribed: false
  }
}`

tomerovadia: Tagging @fongandrew in case he knows more or can help us figure out a way to do this without an API call.

fongandrew: Hmm, do we actually even need the username? We just use this for posting messages saying "So-and-so did a thing" right? In that case, we could probably just do this:

```
<@U017R2YDDAN> did a thing
```

☝️ If you use the `mrkdwn` type for text formatting, Slack will automatically replace the `<@UXXXXXXXX>` with a human readable name, along with a link to see that person's profile.

liewegas: I would love to change all of these to ``<@id>`` but that triggers user notifications that are super annoying.  I tried setting ``link_names`` to false but that didn't seem to help... is there a way to get slack to resolve the ids to names but not trigger notifications?

fongandrew: Oof, yeah, that'll trigger notifications for users. Uh ... in that case, I think we're back to square 1 here.

So, for context, here's what that username field is about: https://api.slack.com/changelog/2017-09-the-one-about-usernames.The username is auto set (usually based on email addresses) and not easy for users to change. That said, it's usually not too hard to figure out who it's referencing. If you try to mention a username, that person should show up in the autocomplete.

If we want names the user actually picked though, you'll need to make the extra API call. We could try caching it though, since these names don't change very often.

PR #166, opened 2020-10-22
clear voter claims when resetting demo

liewegas:
This obscures a complete history in the database, but these are demo
voters--we really don't care.  (And the table doesn't provide a way
to ignore claims pre-reset.)


liewegas:
Tryn ran into this one last night!

tomerovadia (commented):
Is the purpose of this change ease of developing/QA, or something else?

The original use case for demo conversations requires that this `volunteer_voter_claims` history be persisted (for volunteer-to-volunteer practice and admin review of volunteer readiness). The history of voter claims provides a way to determine which volunteers answered which demo conversations, to be matched against their messages and then evaluated to determine if volunteers are prepared to take on real voters.

We can add an `archived` column similarly to what exists for the `messages` table: https://github.com/tomerovadia/VoterHelpLine/pull/79

That also allows us to keep the `volunteer_voter_claims` table write-only.

liewegas:
Much better.  Updated!

src/db_api_util.ts
```diff
@@ -531,6 +531,33 @@ export async function archiveMessagesForDemoVoter(
   }
 }
 
+export async function archiveDemoVoterClaims(
```
tomerovadia: nit: can we name this archiveDemoVolunteerVoterClaims?

src/db_api_util.ts
```diff
@@ -531,6 +531,33 @@ export async function archiveMessagesForDemoVoter(
...
+  userId: string,
+  twilioPhoneNumber: string
+): Promise<void> {
+  logger.info(`ENTERING DBAPIUTIL.clearDemoVoterClaims`);
```
tomerovadia: update log line

src/db_api_util.ts
```diff
@@ -531,6 +531,33 @@ export async function archiveMessagesForDemoVoter(
...
+      SET archived = true
+      WHERE
+        is_demo = true
+        AND user_id = $1
+        AND twilio_phone_number = $2`,
+      [userId, twilioPhoneNumber]
+    );
+
+    logger.info(
+      `DBAPIUTIL.clearDemoVoterClaims: Successfully cleared demo voter claims.`
```
tomerovadia: if you update the function name let's update the log line here too

sql/20201022.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE volunteer_voter_claims
```
tomerovadia: can you update the name of this file to today's date before merging?

tomerovadia (changes requested):
Looks good!

tomerovadia (approved):

PR #167, opened 2020-10-23
add is_demo, routed to threads; add /follow-up command

liewegas:
based on top of https://github.com/tomerovadia/VoterHelpLine/pull/166 so that we don't lose the extra conditional in the WITH sql clause

- we add ``is_demo`` and ``routed`` to threads to avoid consly SQL constructions on the fly
- add ``/follow-up <days>`` command to list voters the volunteer might want to follow up with

I'm a bit worried that in practice this list will get so long for active volunteers that they won't be able to manage if effectively... unless they have a way to permanently "resolve" a voter and take it off their follow-up plate.  I think we should do that by allowing themselves to unclaim the voter.  Will open a separate PR for that.


external_config/slack.md
```diff
@@ -190,3 +190,10 @@ This should be added as a **global** shortcut.
 - **Request URL:** base URL + /slack-command
 - **Short Description:** Broadcast helpline status to channels or volunteers
 - **Usage Hint:** channel-status|volunteer-status [optional preamble message]
+
+#### Follow-up
+
+- **Command:** /follow-up
```
tomerovadia, line 213: can we rename to /needs-follow-up so it's consistent with /needs-attention?

external_config/slack.md
```diff
@@ -190,3 +190,10 @@ This should be added as a **global** shortcut.
...
+- **Request URL:** base URL + /slack-command
+- **Short Description:** List voters to follow-up with
```
tomerovadia, line 215: Can we rename `List voters needing your follow-up` so it's consistent with `/needs-attention`?

src/async_jobs.ts
```diff
@@ -74,6 +73,17 @@ async function slackCommandHandler(
       );
       return;
     }
+    case '/follow-up': {
+      await SlackInteractionHandler.handleCommandFollowUp(
```
tomerovadia, line 77: rename `handleCommandNeedsFollowUp`

src/slack_interaction_handler.ts
```diff
@@ -829,6 +829,65 @@ export async function handleCommandBroadcast(
...
+): Promise<void> {
+  const args = text.split(' ');
+  if (args.length != 1) {
+    await SlackApiUtil.sendEphemeralResponse(
+      responseUrl,
+      `Usage: \`/follow-up <days>`
+    );
+    return;
+  }
+  const days = Number(args[0]);
```
tomerovadia: What happens if they enter a non-number argument accidentally?

src/db_api_util.ts
```diff
@@ -1031,6 +1052,67 @@ export async function getThreadNeedsAttentionFor(
...
+        , s.voter_status
+        FROM threads t, latest_claims c, latest_status s
+        WHERE
+          needs_attention = false
+          AND routed = false
+          AND updated_at <= NOW() - interval '${days} days'
+          AND t.user_id=c.user_id
+          AND c.volunteer_slack_user_id=$1
+          AND s.user_id = t.user_id
+        ORDER BY updated_at`,
```
tomerovadia, line 1123: Should we order by latest_update_age?

src/db_api_util.ts
```diff
@@ -1031,6 +1052,67 @@ export async function getThreadNeedsAttentionFor(
...
+      ), latest_status AS (
+        SELECT user_id, voter_status
+        FROM all_status
+        WHERE rn = 1
+      )
+      SELECT
+        t.slack_parent_message_ts
+        , t.slack_channel_id
+        , t.user_id
+        , EXTRACT(EPOCH FROM now() - updated_at) as last_update_age
```
tomerovadia, line 1110: Is this the last time the voter's thread was updated (turning needs attention on or off)? Can we do the last time the helpline sent them a message instead?

src/db_api_util.ts
```diff
@@ -1031,6 +1052,67 @@ export async function getThreadNeedsAttentionFor(
...
 
+// Return a list of ThreadInfo's for all threads needing followup
+export async function getThreadsNeedingFollowUp(
+  slackUserId: string,
+  days: number
+): Promise<ThreadInfo[]> {
+  const client = await pool.connect();
+  try {
+    const result = await client.query(
+      `WITH all_claims AS (
```
tomerovadia: You can shrink the number of rows you use in the `WITH` clauses significantly if you use `SELECT DISTINCT ON`

src/db_api_util.ts
```diff
@@ -1031,6 +1052,67 @@ export async function getThreadNeedsAttentionFor(
...
+        , c.volunteer_slack_user_name
+        , s.voter_status
+        FROM threads t, latest_claims c, latest_status s
+        WHERE
+          needs_attention = false
+          AND routed = false
+          AND updated_at <= NOW() - interval '${days} days'
+          AND t.user_id=c.user_id
+          AND c.volunteer_slack_user_id=$1
+          AND s.user_id = t.user_id
```
tomerovadia, line 1120: don't you have to match twilio_phone_number = twilio_phone_number as well?

src/db_api_util.ts
```diff
@@ -1031,6 +1052,67 @@ export async function getThreadNeedsAttentionFor(
...
+      SELECT
+        t.slack_parent_message_ts
+        , t.slack_channel_id
+        , t.user_id
+        , EXTRACT(EPOCH FROM now() - updated_at) as last_update_age
+        , c.volunteer_slack_user_name
+        , s.voter_status
+        FROM threads t, latest_claims c, latest_status s
+        WHERE
+          needs_attention = false
```
tomerovadia: why needs_attention false?

src/db_api_util.ts
```diff
@@ -677,6 +680,24 @@ export async function setThreadNeedsAttentionToDb(
   }
 }
 
+export async function setThreadRoutedToDb(
```
tomerovadia: Does `routed` mean routed to a state channel? if so can you rename?

src/db_api_util.ts
```diff
@@ -1031,6 +1052,67 @@ export async function getThreadNeedsAttentionFor(
...
+          , volunteer_slack_user_id
+          , volunteer_slack_user_name
+          , row_number () OVER (PARTITION BY user_id ORDER BY created_at DESC) AS rn
+        FROM volunteer_voter_claims
+        WHERE archived != true
+      ), latest_claims AS (
+        SELECT user_id, volunteer_slack_user_id, volunteer_slack_user_name
+        FROM all_claims
+        WHERE rn = 1
+      ), all_status AS (
```
tomerovadia: can you rename to all_statuses to be consistent with all_claims?

src/db_api_util.ts
```diff
@@ -1031,6 +1052,67 @@ export async function getThreadNeedsAttentionFor(
...
+        t.slack_parent_message_ts
+        , t.slack_channel_id
+        , t.user_id
+        , EXTRACT(EPOCH FROM now() - updated_at) as last_update_age
+        , c.volunteer_slack_user_name
+        , s.voter_status
+        FROM threads t, latest_claims c, latest_status s
+        WHERE
+          needs_attention = false
+          AND routed = false
```
tomerovadia: what's the purpose of filtering out routed? does this mean that I wouldn't be able to see which voters I messaged in the lobby who need follow-up?

external_config/slack.md
liewegas: I didn't want to do `/needs-follow-up` because the you couldn't to `/n` enter enter and take advantage of auto-complete.  This way you can to `/fo` enter enter.  I couldn't think a single-letter command name that took less than 2 chars..

src/db_api_util.ts
liewegas: It's effectively the same, except currently it's ordered with oldest-first... which is I think what we want?

src/db_api_util.ts
liewegas: `updated_at` is only bumped when there is a new message on the thread. It's not affected by automated messages

src/db_api_util.ts
liewegas: oh right!  using is_demo instead (the phone # isn't in the threads table but is_demo is everywhere)

src/db_api_util.ts
liewegas: just thinking it was a distinct set from the attention group, but it doesn't need to be that way.. dropping it!

src/db_api_util.ts
liewegas: It's more like `isThreadRouted` but that `ToDb` suffix is used on a bunch of methods... seems redundant tho given we're in DbApiUtil namespace

src/db_api_util.ts
liewegas: We only want the current active thread for the user, not the lobby or any other earlier iterations of the thread.  The lobby (actually, any new) thread is routed=false initially.

src/slack_interaction_handler.ts
liewegas: fixed this

external_config/slack.md
tomerovadia: sounds good thanks for explaining!

sql/20201027-is-demo.sql
```diff
@@ -0,0 +1,35 @@
+--- add is_demo to threads
+ALTER TABLE threads
+ADD COLUMN is_demo boolean;
+
+WITH demo_number AS (
+    SELECT DISTINCT(to_phone_number) as phone
```
tomerovadia: I believe you're selecting phone numbers here of people who texted with a demo line, right?

Consider a row where the to_phone_number is a person texting the line and the front_phone_number is the demo line

sql/20201027-is-demo.sql
```diff
@@ -0,0 +1,35 @@
...
+    SELECT DISTINCT(slack_channel) FROM messages
+    WHERE
+        EXISTS(SELECT 1 from demo_number WHERE to_phone_number = phone)
+        OR EXISTS(SELECT 1 from demo_number WHERE from_phone_number = phone)
+)
+UPDATE threads
+SET is_demo = false
+WHERE NOT EXISTS (
+    SELECT 1 FROM demo_channel d WHERE threads.slack_channel_id = d.slack_channel
+);
```
tomerovadia: can you add a newline at end of file?

sql/20201027-is-demo.sql
```diff
@@ -0,0 +1,35 @@
...
+        EXISTS(SELECT 1 from demo_number WHERE to_phone_number = phone)
+        OR EXISTS(SELECT 1 from demo_number WHERE from_phone_number = phone)
+)
+UPDATE threads
+SET is_demo = true
+WHERE EXISTS (
+    SELECT 1 FROM demo_channel d WHERE threads.slack_channel_id = d.slack_channel
+);
+
+WITH demo_number AS (
```
tomerovadia: can't these two WITH clauses be one? selecting all slack_channels for is_demo? is_demo is a column in messages

sql/20201027-routed.sql
```diff
@@ -0,0 +1,38 @@
...
+WHERE NOT EXISTS (
+    SELECT 1 FROM threads_with_rank r
+    WHERE
+        r.slack_channel_id = threads.slack_channel_id
+        AND r.slack_parent_message_ts = threads.slack_parent_message_ts
+        AND r.user_id = threads.user_id
+        AND r.is_demo = threads.is_demo
+        AND rn > 1
+)
+;
```
tomerovadia: add newline

src/db_api_util.ts
tomerovadia: can we add twilio_phone_number instead of is_demo to the threads table (or both)? twilio_phone_number and user_id are used throughout the code and in Redis for identifying a unique user, and I'd like to keep that consistent.

src/db_api_util.ts
tomerovadia: remind me: is updated_at updated when a volunteer messages?

src/db_api_util.ts
```diff
@@ -1024,6 +1045,56 @@ export async function getThreadNeedsAttentionFor(
...
+  slackUserId: string,
+  days: number
+): Promise<ThreadInfo[]> {
+  const client = await pool.connect();
+  try {
+    const result = await client.query(
+      `WITH latest_claims AS (
+        SELECT DISTINCT ON (user_id) user_id, volunteer_slack_user_id, volunteer_slack_user_name, is_demo
+        FROM volunteer_voter_claims
+        WHERE archived != true
```
tomerovadia: can we use IS NOT TRUE instead of != true?

src/db_api_util.ts
```diff
@@ -1024,6 +1045,56 @@ export async function getThreadNeedsAttentionFor(
...
+      SELECT
+        t.slack_parent_message_ts
+        , t.slack_channel_id
+        , t.user_id
+        , EXTRACT(EPOCH FROM now() - updated_at) as last_update_age
+        , c.volunteer_slack_user_name
+        , s.voter_status
+        FROM threads t, latest_claims c, latest_statuses s
+        WHERE
+          routed = false
```
tomerovadia: can we use NOT routed instead of routed = false?

tomerovadia: do you want routed = false or true?

src/router.ts
```diff
@@ -622,10 +623,9 @@ const routeVoterToSlackChannel = async (
     userInfo[userInfo.activeChannelId],
     userInfo.activeChannelId
   );
-  await DbApiUtil.setThreadNeedsAttentionToDb(
+  await DbApiUtil.setThreadRoutedToDb(
```
tomerovadia: can you move both this and getThreadNeedsAttentionFor to the very end of this function so it gets to the production-critical stuff first?

src/db_api_util.ts
liewegas: yes.  volunteer text or voter message.

sql/20201027-is-demo.sql
liewegas: there's the direction='INBOUND' condition below?

sql/20201027-is-demo.sql
liewegas: aie.. yes they could, i'm not sure why i didn't notice is_demo in messages.

src/db_api_util.ts
liewegas: false; routed=true threads are dead

src/router.ts
liewegas: We can't move the getThreadNeedsAttentionFor very far because we need the value to set up the new thread.  I can move the setThreadRouted call by getting rid of the early return.

I'm not sure it's worth thinking too hard about what happens if this is interrupted when it pretty much *all* of the code assumes things aren't interrupted.

sql/20201027-routed.sql
tomerovadia: I still see the no newline error

sql/20201027-routed.sql
```diff
@@ -0,0 +1,38 @@
+ALTER TABLE threads
+ADD COLUMN routed boolean;
+
+
+--- populate routed threads
```
tomerovadia: I think I just figured out what `routed` means. Can we rename it? Am I understanding correctly that it means `inactive`? Or is it something else?

sql/20201027-routed.sql
```diff
@@ -0,0 +1,38 @@
...
+WITH threads_with_rank AS (
+    SELECT *, row_number() OVER (PARTITION BY user_id, is_demo ORDER BY updated_at DESC) AS rn
+    FROM threads
+)
+UPDATE threads
+SET routed='t'
```
tomerovadia: can you update this to `routed = true` to be consistent with your other migrations?

sql/20201027-routed.sql
```diff
@@ -0,0 +1,38 @@
...
+        AND rn > 1
+)
+;
+--- false for active threads
+WITH threads_with_rank AS (
+    SELECT *, row_number() OVER (PARTITION BY user_id ORDER BY updated_at DESC) AS rn
+    FROM threads
+)
+UPDATE threads
+SET routed='f'
```
tomerovadia: routed = false for consistency

sql/20201027-routed.sql
```diff
@@ -0,0 +1,38 @@
...
+UPDATE threads
+SET routed='t'
+WHERE EXISTS (
+    SELECT 1 FROM threads_with_rank r
+    WHERE
+        r.slack_channel_id = threads.slack_channel_id
+        AND r.slack_parent_message_ts = threads.slack_parent_message_ts
+        AND r.user_id = threads.user_id
+        AND r.is_demo = threads.is_demo
+        AND rn > 1
```
tomerovadia: If a voter is routed from lobby to channel A to channel B, won't the thread for channel A have rn = 2 and the thread for channel B have rn = 3, but both will end up with routed = true even though channel A's thread is inactive?

maybe this is me still not understanding what routed is. regardless, can you please add more comments about what you're intending to happen in these SQL queries and what routed is supposed to represent, both here and in other files that deal with the routed logic? I want this to be easy to read and understand for future developers.

src/db_api_util.ts
tomerovadia: good point thanks

src/router.ts
```diff
@@ -728,84 +724,89 @@ const routeVoterToSlackChannel = async (
...
-    userInfo,
-    redisClient,
-    twilioPhoneNumber,
-    {
-      destinationSlackChannelName,
-      destinationSlackChannelId,
-      destinationSlackParentMessageTs: userInfo[destinationSlackChannelId],
-    },
-    timestampOfLastMessageInThread
+  await DbApiUtil.setThreadRoutedToDb(
```
tomerovadia: the differ is freaking out here but can you confirm the only change is to move setThreadRoutedToDb lower?

sql/20201027-routed.sql
tomerovadia: I understand it now (missed that you had updated updated_at DESC). Can you add comments explaining that you’re marking all but the latest threads routed (or inactive, ideally).

sql/20201027-routed.sql
liewegas: right :)  yeah, i'll rename

src/router.ts
liewegas: Yep!

liewegas:
Instead of routed -> inactive I did routed -> active and negated the checks to avoid double-negatives like `inactive = false`.

src/db_api_util.ts
```diff
@@ -563,7 +565,14 @@ export async function archiveDemoVoter(
...
       [userId, twilioPhoneNumber]
     );
-
+    await client.query(
+      `UPDATE threads
+      SET archived = true
+      WHERE
+        is_demo = true
+        AND user_id = $1`,
+      [userId]
```
tomerovadia, line 574: can you use twilioPhoneNumber here, to be consistent with the three other queries above? if threads doesn't have twilio_phone_number then it's fine -- we can fix this after the election.

tomerovadia (commented):
I don't have time to look as closely as yesterday, but if you feel good about this please test with Tryn and if all is well in QA (including PR #191 and #193) then we can deploy tonight.

The only major thing that I think might need attention but hopefully not urgently is what we do with threads when a voter is routed back to a channel they've been in. The routing logic returns the voter to their original thread in the channel to which they return. But as far as I can tell we write a new threads row for that same thread, which could create issues if anywhere we expect threads to be unique. If this is indeed an issue, unless it's critical, I'd prefer if we could hold off on fixing it, since I really need to focus my attention on making sure the core functionality is working correctly.

tomerovadia (approved):

PR #168, opened 2020-10-23
add 'Release claim' button

liewegas:
image


src/slack_interaction_handler.ts
```diff
@@ -348,15 +355,17 @@ export async function handleVolunteerUpdate({
...
+      : `*Operator:* Volunteer claim released by *${originatingSlackUserName}* at *${specialSlackTimestamp}*.`,
     {
       parentMessageTs: payload.container.thread_ts,
       channel: payload.channel.id,
     }
   );
 
   logger.info(
-    `SLACKINTERACTIONHANDLER.handleVoterStatusUpdate: Successfully sent message recording voter status change`
+    `SLACKINTERACTIONHANDLER.handleVolunteerUpdate: Successfully sent message recording volunteer status change`
```
tomerovadia: nit: can we do "volunteer claim" instead of "volunteer status"?

src/slack_interaction_handler.ts
```diff
@@ -348,15 +355,17 @@ export async function handleVolunteerUpdate({
 
   // Post a message in the voter thread recording this status change.
   await SlackApiUtil.sendMessage(
-    `*Operator:* Volunteer changed to *${selectedVolunteerSlackUserName}* by *${originatingSlackUserName}* at *${specialSlackTimestamp}*.`,
+    selectedVolunteerSlackUserName
+      ? `*Operator:* Volunteer changed to *${selectedVolunteerSlackUserName}* by *${originatingSlackUserName}* at *${specialSlackTimestamp}*.`
+      : `*Operator:* Volunteer claim released by *${originatingSlackUserName}* at *${specialSlackTimestamp}*.`,
```
tomerovadia, line 360: Is `originatingSlackUserName` null here in the release claim case?

liewegas: no, originatingSlackUserName is the person who clicked the button (not necessarily the old assigned volunteer)

liewegas: I wasn't certain if the payload's value for the drop-down would reliably have the old volunteer so i left it off the message. I'm not sure it is that important.. you can look in the history to see who the assigned person was before

tomerovadia: Right thanks! No problem, I don't think there's a need to have the message include the old volunteer claim.

src/slack_block_util.ts
```diff
@@ -136,6 +136,17 @@ const volunteerSelectionPanel: SlackBlock = {
         emoji: true,
       },
     },
+    {
+      type: 'button',
+      style: 'primary',
+      text: {
+        type: 'plain_text',
+        text: 'Release claim',
```
tomerovadia: I wonder if "Clear volunteer" is clearer for the button text

tomerovadia (commented):
Can you add a screenshot to the description?

Also can you ask if Andrew can review? Feel free to QA in the meantime.

src/slack_block_util.ts
liewegas: updated!

tomerovadia (approved):

PR #173, opened 2020-10-24
handle VOTED after welcome message (when userInfo but not thread)

liewegas:


src/router.ts
```diff
@@ -945,8 +945,27 @@ export async function clarifyHelplineRequest(
   const userInfo = userOptions.userInfo;
   userInfo.lastVoterMessageSecsFromEpoch = Math.round(Date.now() / 1000);
 
+  let messageToVoter = MessageConstants.CLARIFY_HELPLINE_REQUEST();
+
+  if (isVotedKeyword(userOptions.userMessage)) {
+    messageToVoter = MessageConstants.VOTED_RESPONSE();
```
tomerovadia: Can't this use VOTED_WELCOME_RESPONSE? That way they know they still have options to enter the helpline or stop if needed.

liewegas: Yeah, updated.  I was trying to be less repetitive with the STOP and Msg&data rates boilerplate but it occurs to me now that there may be a long delay between the two.

tomerovadia: Ah good point.

tomerovadia (approved):

PR #179, opened 2020-10-25
no VOTED after HELPLINE

liewegas:


src/message_constants.ts
```diff
@@ -72,7 +72,7 @@ export function CLARIFY_HELPLINE_REQUEST(): string {
 
 // Only used by VOTE_AMERICA.
 export function WELCOME_AND_STATE_QUESTION(): string {
-  return "Welcome to VoteAmerica! Msg&data rates may apply.\n\nReply VOTED if you've already voted, STOP to unsubscribe.\n\nIn which U.S. state are you looking to vote?";
+  return "Welcome to the VoteAmerica Helpline! Msg&data rates may apply. Reply STOP to unsubscribe.\n\nIn which U.S. state are you looking to vote?";
```
tomerovadia: Can we change to ```Welcome to the VoteAmerica Helpline! In which U.S. state are you looking to vote?\n\nMsg&data rates may apply. Reply STOP to unsubscribe.``` so that hopefully more people see that we're asking for a U.S. state and don't linger in the lobby? I got Brenna's approval

tomerovadia (commented):
Looks like checks are failing

tomerovadia (approved):

PR #180, opened 2020-10-25
post attachments to #attachments channel first

liewegas:
This is a slack workaround so that the bot can make the images public.


fongandrew (approved):

liewegas:
@fongandrew I had to add a sleep in there to make this work... seems not ideal, but it's working.

src/slack_api_util.ts
```diff
@@ -72,10 +72,35 @@ export async function getThreadPermalink(
...
+      errors.push(`Unable to post permalink to attachments channel`);
+      logger.error(
+        `SLACKAPIUTIL.makeFilesPublic: unable to post permalink to channel: ${JSON.stringify(
+          response.data
+        )}`
+      );
+      continue;
+    }
+
+    await delay(2000);
```
benweissmann, line 100: Do we know why this is needed / how reliable this is? Maybe we need to wait for the message-posted even from the slack channel and then we can do this? Or maybe we have to wait until the unfurl happens (not sure if there's an event for that...). What happens if this delay is too short?

liewegas: If it doesn't work you get a red X, so there is feedback at least.  It's hard to experiment because I've only reproduced this in staging.  I think i need to set up a separate slack user locally to make it happen there.

My take is we merge this now since it gets it working, and follow up with something more elegant.

liewegas: I really wish there was a solution that didn't result in a Slackbot ping about the attachment being shared.. it's kind of annoying.

PR #181, opened 2020-10-25
Update how STOP keyword is detected

tomerovadia:
In particular, loosen rule to include two additional circumstances:
1) STOP with punctuation
2) STOP in all caps in context of a longer message

### DIFF
--- a/src/keyword_parser.test.js
+++ b/src/keyword_parser.test.js
@@ -1,0 +1,38 @@
+const KeywordParser = require('./keyword_parser');
+
+test('Does not consider a normal message to contain STOP keyword.', () => {
+  expect(KeywordParser.containsStopKeyword('Some normal message')).toBe(false);
+});
+
+test('Detects STOP keyword.', () => {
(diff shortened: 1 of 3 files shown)


src/keyword_parser.test.js
```diff
@@ -0,0 +1,50 @@
...
+    KeywordParser.containsStopKeyword(
+      "I'll stop by the polling place on the way to work"
+    )
+  ).toBe(false);
+});
+
+test('Detects STOP keyword in the context of longer message if in all uppercase.', () => {
+  expect(
+    KeywordParser.containsStopKeyword(
+      "That's what I am telling you, I have done what you told me to do to unsubscribe and YOU ARE STILL CONTACTING ME. STOP IT!!"
```
benweissmann, line 47: hmm... do we have a sense of how common this is? I'm worried that some people just type in caps a lot and we're going to get false-positives. Maybe we try without this and see if we're still getting a lot of opt-outs going through?

tomerovadia: sounds good! should I split this PR up so that we can at least catch "STOP!!!!"

PR #182, opened 2020-10-25
Add retries to Slack API requests (which are issued via axios)

tomerovadia:
Retry 3 times, 2000 ms apart.

This affects:
1. relaying of voter messages to Slack
2. relaying of operator messages to Slack, including message history
3. all other requests issued to the Slack API.


src/slack_interaction_api_util.ts
```diff
@@ -1,11 +1,19 @@
 import axios from 'axios';
+import axiosRetry from 'axios-retry';
 import { voterStatusPanel, SlackBlock } from './slack_block_util';
 import logger from './logger';
 import { UserInfo } from './types';
 import * as SlackApiUtil from './slack_api_util';
 import * as SlackInteractionHandler from './slack_interaction_handler';
 import { PromisifiedRedisClient } from './redis_client';
 
+axiosRetry(axios, {
```
benweissmann: Can we just import the `slackAPI` client from the other file and use that? Should also simplify all the axios calls in this file

src/slack_api_util.ts
```diff
@@ -17,6 +18,13 @@ slackAPI.defaults.headers.post[
   'Authorization'
 ] = `Bearer ${process.env.SLACK_BOT_ACCESS_TOKEN}`;
 
+axiosRetry(slackAPI, {
+  // This function is passed a retryCount which can be used to
+  // define custom retry delays.
+  retryDelay: () => 1000 /* millisecs between retries*/,
```
benweissmann: Your PR description says 2000ms but this is 1000ms -- which one did you intend?

benweissmann (changes requested):

src/slack_api_util.ts
tomerovadia: Good catch, thanks! Changed to 2000

src/slack_interaction_api_util.ts
tomerovadia: Yep! Did I do this correctly?:

1) Will the export include the modifications to slackAPI util in-file (e.g. retries)?
2) Did I handle params nesting for slackAPI methods correctly?
3) Did I preserve get/post methods correctly?

benweissmann (approved):
looks great!

PR #183, opened 2020-10-25
Update STOP keyword detection to not consider punctuation

tomerovadia:
Loosen rules to detect STOP sent with punctuation

### DIFF
--- a/src/keyword_parser.test.js
+++ b/src/keyword_parser.test.js
@@ -13,9 +13,9 @@ test('Detects STOP keyword with spaces on either side.', () => {
 });
 
 test('Does not detect STOP keyword with spaces within the word.', () => {
-  expect(
-    KeywordParser.isStopKeyword('VOTING IS TOP OF MIND FOR ME YAY')
-  ).toBe(false);
+  expect(KeywordParser.isStopKeyword('VOTING IS TOP OF MIND FOR ME YAY')).toBe(
+    false
+  );
 });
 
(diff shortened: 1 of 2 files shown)


liewegas:
This is fine, but keep in mind that for VA at least the muxer is already translating a lot of things into just ``STOP`` for you.

See https://github.com/vote/twilio-webhook-muxer/blob/master/serverless.yml#L132 for the list, and https://github.com/vote/twilio-webhook-muxer/blob/master/app/muxer.py#L64 for the normalization.

tomerovadia:
> This is fine, but keep in mind that for VA at least the muxer is already translating a lot of things into just `STOP` for you.
> 
> See https://github.com/vote/twilio-webhook-muxer/blob/master/serverless.yml#L132 for the list, and https://github.com/vote/twilio-webhook-muxer/blob/master/app/muxer.py#L64 for the normalization.

Got it. If it's already handling cases like `STOP!!!!` then we can forgo this PR. I thought I noticed a few examples though where that was not caught (I can find them again if needed).

liewegas:
It won't catch ``foo bar stop!!!!`` but it will catch ``....STOP...!!!...``

tomerovadia:
> It won't catch `foo bar stop!!!!` but it will catch `....STOP...!!!...`

cool that's what @benweissmann wants for now. We can skip this PR for now then. Let me see if I can find an example where it wasn't caught -- it might have included something else.

PR #184, opened 2020-10-26
Have Redis update sooner when a voter is routed between channels

tomerovadia:
Currently, a number of operations execute between a) the time a new thread is created in a new Slack channel to which a voter is being routed, and b) the time that Redis is updated to reflect the voter's new thread and channel:

1. A new row is logged to the DB threads table
2. The message history is posted in the new thread.

In this time, the voter might message again, and if Redis does not reflect the new channel, the voter's message might trigger another re-routing of the voter.

We saw this happen twice today (the voter texted two consecutive texts within 1-2 seconds).

This PR moves the Redis update happen before the two operations listed above, so that the chance increases that the voter's subsequent message will successfully be sent to the new channel/thread.

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -480,6 +480,24 @@ const routeVoterToSlackChannelHelper = async (
                 destinationSlackParentMessageTs: ${destinationSlackParentMessageTs},
                 destinationSlackChannelName: ${destinationSlackChannelName}`);
 
+  logger.debug(
+    "ROUTER.routeVoterToSlackChannelHelper: Changing voter's active channel."
+  );
+  // Reassign the active channel so that the next voter messages go to the
+  // new active channel.
+  userInfo.activeChannelId = destinationSlackChannelId;
+  userInfo.activeChannelName = destinationSlackChannelName;
+
+  // Update userInfo in Redis (remember state channel thread identifying info and new activeChannel).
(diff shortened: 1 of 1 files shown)


liewegas:
LGTM.

I wonder if we should consider a more complete solution for this class of problem, though, that serializes all processing by user?  It would be a bit tricky to implement since on the slack side it's more complicated than just wrapping the endpoint enter/exit (since there is async work on the modals etc), but it would avoid this entire class of bugs.

benweissmann (approved):
Looks good. I agree with @liewegas about coming up with a more general solution to this problem, but I think that's a post-election thing. If you're OK being locked in to AWS, I've had great success with SQS FIFO queues -- basically, you'd have the endpoints that receive webhooks from twilio/slack just write to the SQS queue, using the user ID as a message group id, and then you do all of the work in lambda functions that read from those queues. Using the message group ids with a FIFO queue guarantees that only one message from each user is being processed at a time, so you don't need to worry about stuff like this (and you also get some other nice features, like automatic retries and deduplication). If you're interested, let's talk more after the election! Would definitely require ditching Heroku for AWS, tho.

tomerovadia:
Agreed! And yes, let's discuss after the election

PR #186, opened 2020-10-27
fix block updates

liewegas:
Regression in merge of 4d1fbe8dc11f8b4bfde2786096d3bc8450c79e9b

### DIFF
--- a/src/slack_interaction_api_util.ts
+++ b/src/slack_interaction_api_util.ts
@@ -17,14 +17,11 @@ export async function replaceSlackMessageBlocks({
   logger.info('ENTERING SLACKINTERACTIONAPIUTIL.replaceSlackMessageBlocks');
   // Replace voter status panel with message.
   const response = await SlackApiUtil.slackAPI.post('chat.update', {
-    params: {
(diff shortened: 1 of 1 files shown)


src/slack_interaction_api_util.ts
```diff
@@ -17,14 +17,12 @@ export async function replaceSlackMessageBlocks({
   logger.info('ENTERING SLACKINTERACTIONAPIUTIL.replaceSlackMessageBlocks');
   // Replace voter status panel with message.
   const response = await SlackApiUtil.slackAPI.post('chat.update', {
-    params: {
-      channel: slackChannelId,
-      token: process.env.SLACK_BOT_ACCESS_TOKEN,
-      ts: slackParentMessageTs,
-      blocks: newBlocks,
-    },
+    'Content-type': 'application/json',
```
tomerovadia: application/json is built into slackAPI here: https://github.com/tomerovadia/VoterHelpLine/blob/master/src/slack_api_util.ts#L17

so you can omit that

tomerovadia (approved):
Thanks for fixing this!

PR #187, opened 2020-10-27
fix WHERE archived != true instances

liewegas:
This does not include NULL rows, which we want


tomerovadia (changes requested):
nit: can we use `IS FALSE`?

tomerovadia (approved):

PR #188, opened 2020-10-27
do not mix demo and non-demo history

liewegas:


tomerovadia (commented):
good catch. can we do twilio_phone_number instead of is_demo?

tomerovadia (commented):
good catch. can we do twilio_phone_number instead of is_demo?

tomerovadia (approved):

PR #190, opened 2020-10-28
make deep copy before modifying initial blocks with already voted status

liewegas:


tomerovadia:
Not sure I follow why we need a copy

tomerovadia (commented):
can we deep clone the slackBlocks within `getVoterStatusBlocks` instead? That way anywhere it's used this issue will be addressed

tomerovadia (approved):

PR #191, opened 2020-10-28
Format Twilio attachments as block 

fongandrew:
This switches incoming Twilio messages to blocks instead of standard Mrkdwn text, and displays any media attachments as Slack button blocks to distinguish them from possibly unsafe links.

Screen Shot 2020-10-27 at 11 27 33 PM

It requires altering the DB with a new JSON column with the media links since we're now storing those as its own data rather than just extra text appended to the message.

### DIFF
--- a/sql/20201027-twilio-attachments.sql
+++ b/sql/20201027-twilio-attachments.sql
@@ -1,0 +1,5 @@
+-- Store the array of slack_attachments as a jsonb column. We use jsonb rather than
+-- a text array because they have similar performance, but JSON is easier to
+-- sync to a Redshift data warehouse (e.g. Civis)
+ALTER TABLE messages
+ADD COLUMN twilio_attachments jsonb;
--- a/src/db_api_util.ts
+++ b/src/db_api_util.ts
@@ -158,7 +158,7 @@ export async function logMessageToDb(
         databaseMessageEntry.archived,
         databaseMessageEntry.stateName,
 
-        // We can't just pass these arrays as-is because this are JSON columns,
+        // We can't just pass these arrays as-is because these are JSON columns,
         // and node-pg serializes arrays as PG arrays rather than JSON arrays.
         databaseMessageEntry.slackFiles
           ? JSON.stringify(databaseMessageEntry.slackFiles)
--- a/src/db_api_util.ts
+++ b/src/db_api_util.ts
@@ -44,6 +44,7 @@ export type DatabaseMessageEntry = {
   archived?: boolean | null;
   stateName?: string | null;
   slackFiles?: SlackFile[] | null;
+  twilioAttachments?: string[] | null;
 };
 
 export type DatabaseVoterStatusEntry = {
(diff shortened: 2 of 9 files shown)


src/slack_message_formatter.ts
```diff
@@ -1,11 +1,24 @@
...
     .map((x) => '>' + x)
+    .concat(
+      attachments?.length
+        ? [
+            `*Attachments:* ${attachments
+              .map((url, i) => `<${url}|Attachment ${i + 1}>`)
+              .join(' ')}`,
+          ]
+        : []
+    )
```
fongandrew, line 21: I considered displaying the message history using blocks so I could intersperse buttons for attachments as needed, but it ended up taking more vertical space and making things unnecessarily complicated for an edge case, so I decided to just stick with plain ol' links in this case. I imagine the need to open attachments from before routing will be rare.

Screen Shot 2020-10-28 at 12 51 15 AM

sql/20201027-twilio-attachments.sql
```diff
@@ -0,0 +1,5 @@
+-- Store the array of slack_attachments as a jsonb column. We use jsonb rather than
+-- a text array because they have similar performance, but JSON is easier to
```
tomerovadia, line 2: Do you know if Redshift has a JSON type? I couldn't make that work this morning. I used text instead per a suggestion I saw somewhere

src/twilio_util.ts
```diff
@@ -37,25 +37,23 @@ export type TwilioRequestBody = {
   [mediaKey: string]: string | undefined;
 };
 
-export function formatAttachments(reqBody: TwilioRequestBody): string {
+/** Returns list of URLs for MMS attachments */
```
tomerovadia: Let's use // for comments, for consistency

src/twilio_util.ts
```diff
@@ -37,25 +37,23 @@ export type TwilioRequestBody = {
...
+export function getAttachments(reqBody: TwilioRequestBody): string[] {
   const numMedia = Number(reqBody.NumMedia);
 
   if (numMedia === 0) {
```
tomerovadia: since this is going to run for every single received message, can we only do the Number() conversion if !== 0? In other words, check numMedia === "0" as a string so that we can skip the conversion for 99.9% of messages

src/db_api_util.ts
```diff
@@ -157,11 +158,14 @@ export async function logMessageToDb(
         databaseMessageEntry.archived,
         databaseMessageEntry.stateName,
 
-        // We can't just pass the array as-is because this is a JSON column,
+        // We can't just pass these arrays as-is because this are JSON columns,
```
tomerovadia: typo ("this are")

src/slack_message_formatter.ts
tomerovadia: This is absolutely fine! Thanks for thinking of this

src/router.ts
```diff
@@ -839,10 +853,15 @@ export async function determineVoterState(
   logger.debug(
     `ROUTER.determineVoterState: Passing voter message to Slack, slackChannelName: ${lobbyChannelId}, parentMessageTs: ${lobbyParentMessageTs}.`
   );
+  const blocks = SlackBlockUtil.formatMessageWithAttachmentLinks(
```
tomerovadia, line 856: is there any way to have this be part of SlackApiUtil.sendMessage, so that it's only called once instead of needing it in every router function?

sql/20201027-twilio-attachments.sql
fongandrew: I just copied this from https://github.com/tomerovadia/VoterHelpLine/blob/master/sql/20201023-mms.sql actually. Maybe @liewegas can comment.

src/twilio_util.ts
fongandrew: Sure -- although for reference, JSDoc syntax is nice because TypeScript and VSCode will parse this and use it for autocomplete purposes.

src/router.ts
fongandrew: We could, although I think it's mixing concerns a bit. For reference, we call `SlackBlockUtil.formatMessageWithAttachmentLinks` 4 times in this file and `SlackApiUtil.sendMessage` 23 times. The other thing to note is that we're not just adding some attachments in `formatMessageWithAttachmentLinks`. It treats the underlying text format differently. In `sendMessage`, we assume the `message` is Mrkdwn and format accordingly. But `formatMessageWithAttachmentLinks` escapes all of that because it knows the text is coming from random people over Twilio. So we could move that logic into `sendMessage`, but it's a bit more than just adding an `attachments` param.

tomerovadia: That absolutely makes sense. Thanks for giving this a lot of thought.

src/twilio_util.ts
tomerovadia: Oh that's very good to know. We should change all the comments in that case. Let's do that after the election. Thanks for pointing that out.

tomerovadia (approved):

PR #192, opened 2020-10-28
Revamp needs-attention handling

liewegas:
This changes the way we update needs-attention:

- Needs attention no longer depends on whether a volunteer is assigned.
- Needs attention is set on a new thread, and when an incoming message is handled from a cleared voter (i.e., past lobby, or volunteer engaged)
- Needs attention is cleared when a slack thread message is relayed to the voter (i.e., not a command)

Implementation notes:
- We pull the updates out of the twilio/slack send functions, which are deep in the call chain and kludgey.  Instead, we update from high level methods.
- The thread ``updated_at`` field is now updated from ``logMessageToDb``.
- The old method that updates from the databaseMessageEntry is gone; instead the callers explicitly set true or false for the thread.

This is simpler and easier to understand.  The main net change from a volunteer perspective is that if you start talking to a voter that is unclaimed their needs-attention status will get cleared.  However, they will still be unclaimed, so I don't think they will get lost.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -196,6 +196,7 @@ export async function handleKnownVoterBlockLogic(
           parentMessageTs: userInfo[userInfo.activeChannelId],
           channel: userInfo.activeChannelId,
           isVoterMessage: true,
+          isBlocked: true,
         },
         inboundDbMessageEntry,
         userInfo
--- a/src/slack_api_util.ts
+++ b/src/slack_api_util.ts
@@ -40,6 +40,7 @@ type SlackSendMessageOptions = {
   blocks?: SlackBlock[];
   isVoterMessage?: boolean;
   isAutomatedMessage?: boolean;
+  isBlocked?: boolean;
 };
 
 type SlackChannelNamesAndIds = {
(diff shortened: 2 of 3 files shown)


src/app.ts
```diff
@@ -196,6 +196,7 @@ export async function handleKnownVoterBlockLogic(
           parentMessageTs: userInfo[userInfo.activeChannelId],
           channel: userInfo.activeChannelId,
           isVoterMessage: true,
+          isBlocked: true,
         },
         inboundDbMessageEntry,
         userInfo
```
tomerovadia, line 202: I'd like to keep SlackApiUtil.sendMessage as simple as possible. Can we move the updateThreadStatusFromMessage logic here (and elsewhere if needed) instead of passing an isBlocked?

src/slack_api_util.ts
```diff
@@ -239,7 +240,9 @@ export async function sendMessage(
 
       try {
         await DbApiUtil.logMessageToDb(databaseMessageEntry);
-        await DbApiUtil.updateThreadStatusFromMessage(databaseMessageEntry);
```
tomerovadia, line 242: Question about this earlier implementation: SlackApiUtil.sendMessage is used for sending non-voter messages to Slack as well. If a volunteer dropdown is changed, for example, an update is sent to the thread. In those cases, is the thread status updated?

liewegas: Yeah, good point.  I think this needs a bigger clean-up.

PR #193, opened 2020-10-29
fix message history after routing

liewegas:
I'm not sure why userInfo.twilioPhoneNumber isn't set here

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -540,7 +540,7 @@ const routeVoterToSlackChannelHelper = async (
   // Populate state channel thread with message history so far.
   await postUserMessageHistoryToSlack(
     userInfo.userId,
-    userInfo.twilioPhoneNumber,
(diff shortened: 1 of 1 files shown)


tomerovadia (commented):
Can you please share more context? Is something broken in production?

liewegas:
This was broken by c7c56891c4d8c98aaa0d0c7795b8e3d747cdc286 which assumed that userInfo.twilioPhoneNumber would be defined (it's not).  There is a local twilioPhoneNumber value available so use that instead.

It was just recently merged and is not deployed anywhere yet

tomerovadia:
Thanks for explaining! The twilioPhoneNumber is part of the key that is used to derive the userInfo (along with userId), so anywhere in the code that retrieves userInfo should already have the twilioPhoneNumber available (with 1 minor exception).

PR #198, opened 2020-11-01
fix broadcast bad usage error message (and make it ephemeral)

liewegas:
The old error path was just broken.. it was posting a permanent message to the wrong channel (we want all responses to this command to be ephemeral). Also the prod deployment was misconfigured; update the docs to reflect the escaping option for slash commands!

### DIFF
--- a/src/slack_interaction_handler.ts
+++ b/src/slack_interaction_handler.ts
@@ -464,9 +464,10 @@ export async function handleCommandUnclaimed(
       arg = arg.substr(1); // strip off the # prefix
     }
     if (!(arg in slackChannelIds)) {
-      await SlackApiUtil.sendMessage(`Channel #${arg} not found`, {
-        channel: channelId,
-      });
+      await SlackApiUtil.sendEphemeralResponse(
+        responseUrl,
+        `Channel #${arg} not found`
+      );
       return;
     }
     showChannelName = arg;
--- a/external_config/slack.md
+++ b/external_config/slack.md
@@ -193,6 +193,7 @@ This should be added as a **global** shortcut.
 - **Request URL:** base URL + /slack-command
 - **Short Description:** List unclaimed voters in a channel
 - **Usage Hint:** (None)
+- **Escape channels, users, and links sent to your app:** Unchecked
 
 #### Needs attention
 
@@ -200,6 +201,7 @@ This should be added as a **global** shortcut.
 - **Request URL:** base URL + /slack-command
 - **Short Description:** List voters needing your attention
 - **Usage Hint:** (None)
+- **Escape channels, users, and links sent to your app:** Checked
 
 #### Broadcast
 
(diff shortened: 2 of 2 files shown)


src/slack_interaction_handler.ts
```diff
@@ -464,9 +464,10 @@ export async function handleCommandUnclaimed(
       arg = arg.substr(1); // strip off the # prefix
     }
     if (!(arg in slackChannelIds)) {
-      await SlackApiUtil.sendMessage(`Channel #${arg} not found`, {
-        channel: channelId,
-      });
+      await SlackApiUtil.sendEphemeralResponse(
```
tomerovadia, line 467: in addition to the fix, is this changing from a regular message to an ephemeral message? if so can you add this to the PR title and/or description?

src/slack_interaction_handler.ts
```diff
@@ -464,9 +464,10 @@ export async function handleCommandUnclaimed(
       arg = arg.substr(1); // strip off the # prefix
     }
     if (!(arg in slackChannelIds)) {
-      await SlackApiUtil.sendMessage(`Channel #${arg} not found`, {
-        channel: channelId,
```
tomerovadia, line 468: wasn't the issue also that channelId was actually a slack user id in one case? did you figure out how that happened? and can you change the variable name, or before assigning it ensure that it is a channel id?

liewegas: I've lost the original thread with the error message .. do you have a link?

tomerovadia: https://voteamericahelpline.slack.com/archives/G01DNHJ3AM7/p1604253054010300?thread_ts=1604250068.007300&cid=G01DNHJ3AM7

liewegas: It looks like (1) they passed an `@someuser` argument, and (2) slack is configured to escape the arguments.. for this command it should not be. updated the slack config doc to reflect that.

tomerovadia: So there were multiple issues here -- misconfigured slack but also code bugs?

liewegas: user error + misconfigured slack.  this change is fine

tomerovadia: Is there a way to make the code resilient to the user error?

tomerovadia: I'm going to approve this -- if it's still not resilient to user error can you fix or create an issue so we can get back to it at some point?

tomerovadia: (And sorry if it is and I'm not following correctly)

tomerovadia (approved):

src/slack_interaction_handler.ts
liewegas: I think we're good.  The error message makes it clear we expected a channel name, and turning off escaping will make it not look weird.

PR #204, opened 2020-11-10
automatically route voter when phone number state is known

liewegas:


src/message_constants.ts
```diff
@@ -75,6 +75,10 @@ export function WELCOME_AND_STATE_QUESTION(): string {
   return 'Welcome to the VoteAmerica Helpline! In which U.S. state are you looking to vote?\n\nMsg&data rates may apply. Reply STOP to unsubscribe.';
 }
 
+export function WELCOME_FINDING_VOLUNTEER(): string {
```
tomerovadia, line 78: I believe all other initial messages to voters have the msg&data rates part as well as the STOP keyword instructions. Are we okay with this one not having that?

src/router.ts
```diff
@@ -394,35 +411,66 @@ export async function handleNewVoter(
...
-      `ROUTER.handleNewVoter (${userInfo.userId}): LoadBalancer returned Slack channel ${selectedChannelName} for new PUSH voter.`
-    );
-    if (selectedChannelName) {
-      slackChannelName = selectedChannelName;
-    } else {
-      // If LoadBalancer didn't find a Slack channel, then select #national or #demo-national as fallback.
-      slackChannelName = userInfo.isDemo ? 'demo-national' : 'national';
-      logger.error(
-        `ROUTER.handleNewVoter (${userInfo.userId}): ERROR LoadBalancer did not find a Slack channel for new PUSH voter. Using ${slackChannelName} as fallback.`
+  // do we already know the voter's state?
```
tomerovadia: can we do this only for VOTE_AMERICA?

src/router.ts
```diff
@@ -394,35 +411,66 @@ export async function handleNewVoter(
...
+  const knownState = await DbApiUtil.getKnownPhoneState(
+    userOptions.userPhoneNumber
+  );
+  if (knownState) {
```
tomerovadia: If there is a knownState why do we call determineState?

src/router.ts
```diff
@@ -394,35 +411,66 @@ export async function handleNewVoter(
...
+    const stateName = StateParser.determineState(knownState);
+    if (stateName) {
```
tomerovadia: Can you log something if there's a knownState but it's not parsed successfully?

src/router.ts
```diff
@@ -394,35 +411,66 @@ export async function handleNewVoter(
...
+      userInfo.stateName = stateName;
+      slackChannelName = await LoadBalancer.selectSlackChannel(
```
tomerovadia: you're missing a `const`

src/router.ts
```diff
@@ -394,35 +411,66 @@ export async function handleNewVoter(
...
+        redisClient,
+        LoadBalancer.PULL_ENTRY_POINT,
+        stateName,
+        userInfo.isDemo
       );
+      if (slackChannelName) {
+        logger.info(
+          `ROUTER.handleNewVoter (${userInfo.userId}): New voter is in known state {stateName}, channel {slackChannelName}`
```
tomerovadia: instead of `channel {slackChannelName}` can we say something like `selected Slack channel is {slackChannelName}`

src/router.ts
```diff
@@ -394,35 +411,66 @@ export async function handleNewVoter(
...
+      } else {
+        slackChannelName = userInfo.isDemo ? 'demo-national' : 'national';
+        logger.info(
+          `ROUTER.handleNewVoter (${userInfo.userId}): New voter is in known state {stateName}, but no channel match`
+        );
+      }
+    }
+  }
+
+  if (!slackChannelName) {
```
tomerovadia, line 450: we check if slackChannelName is null twice (430 and 443). can we do it once?

src/db_api_util.ts
```diff
@@ -1199,6 +1199,21 @@ export async function getVoterHasVolunteer(userId: string): Promise<boolean> {
...
+export async function getKnownPhoneState(
+  userPhoneNumber: string
+): Promise<string | null> {
+  const client = await pool.connect();
+  try {
+    const result = await client.query(
+      'SELECT state FROM known_phone_states WHERE phone_number = $1',
+      [userPhoneNumber]
+    );
+    return result.rows.length > 0 ? result.rows[0]['state'] : null;
```
tomerovadia, line 1309: if there aren't any rows will rows[0] be null and this throw an error?

src/router.ts
liewegas: knownState will be ``MI`` but helpline wants ``Michigan``.

src/db_api_util.ts
liewegas: we check ``result.rows.length > 0 ``

src/message_constants.ts
liewegas: fixed!

src/router.ts
liewegas: it's a `let` on line 414

src/router.ts
liewegas: i think they're both needed without very awkward code structure

src/router.ts
tomerovadia: Can we change the database table to state names so we don't have to do this parsing online?

liewegas: We could.. I kind of like it this way though because it adds a layer of protection against bad/bogus data in the table.

tomerovadia: Ok no problem whichever you prefer SGTM

src/router.ts
tomerovadia: We'd only enter this `if` block if we didn't find a known state (or didn't try if not VOTE_AMERICA), right? Can you add comments to make it easier for future developers to follow this logic?

src/router.ts
```diff
@@ -206,13 +211,25 @@ const introduceNewVoterToSlackChannel = async (
...
+      if (userInfo.stateName) {
+        messageToVoter = MessageConstants.WELCOME_FINDING_VOLUNTEER();
+      } else {
+        messageToVoter = MessageConstants.WELCOME_AND_STATE_QUESTION();
+      }
     } else {
       // Voter has already received automated welcome and is just now responding with "HELPLINE".
-      messageToVoter = MessageConstants.STATE_QUESTION();
+      if (userInfo.stateName) {
+        messageToVoter = MessageConstants.NO_STATE_FINDING_VOLUNTEER();
```
tomerovadia: Even though the message text might make sense in this case, usage of the message constant in this case is confusing because the state is known but the message constant is NO_STATE_FINDING_VOLUNTEER. Can you clean that up (e.g. rename NO_STATE_FINDING_VOLUNTEER to just FINDING_VOLUNTEER)?

src/router.ts
```diff
@@ -206,13 +211,25 @@ const introduceNewVoterToSlackChannel = async (
...
+      } else {
+        messageToVoter = MessageConstants.STATE_QUESTION();
+      }
     }
   } else {
-    messageToVoter = MessageConstants.WELCOME_VOTER();
+    if (userInfo.stateName) {
```
tomerovadia: This else branch is for non-VOTE_AMERICA right? Would userInfo.stateName ever be already populated at this point for other clients?

liewegas: I see at least one case where it might be? https://github.com/tomerovadia/VoterHelpLine/blob/master/src/router.ts#L404  I don't really understand what the PUSH/PULL stuff is all about or how other orgs are using it.

tomerovadia: Ya sorry I know the PUSH/PULL is confusing. When handleNewVoter is called with a PUSH entrypoint, the state is determined but no message is sent to the voter (below, the `TwilioApiUtil.sendMessage` is guarded by `entryPoint === LoadBalancer.PULL_ENTRY_POINT`.

So basically, the state will only be determined in PUSH entrypoint cases, and a message to voter will only happen in PULL entrypoint cases. So you can leave this as only `messageToVoter = MessageConstants.WELCOME_VOTER();` because messageToVoter will only be used `userInfo.stateName` is null anyway.

Hope that makes sense, but feel free to ask me to clarify more if needed!

src/router.ts
tomerovadia: Did you forget a commit for this one?

tomerovadia (commented):
Mostly LGTM. I think what remains on this one is:
-adding comments for where the branching gets a bit spaghetti-y
-removing the changes to the non-VoteAmerica branch for messageToVoter
-QA

src/router.ts
liewegas: Makes sense!

liewegas:
Added a few comments and fixed the non-VA path.  Tryn already tested this one!

src/router.ts
```diff
@@ -416,8 +416,10 @@ export async function handleNewVoter(
       userOptions.userPhoneNumber
     );
     if (knownState) {
+      // the knownState value from the db is probably a state abbreviation, so we need to resolve it
```
tomerovadia: nit: Can you please capitalize the first first word of each of these comments to be consistent with the rest of the codebase?

tomerovadia (changes requested):

src/router.ts
liewegas: fixed!

tomerovadia (approved):

PR #205, opened 2020-11-10
simplify VOTED status

liewegas:
- ALREADY_VOTED -> VOTED converge
- simplify UI to a drop-down that includes the VOTED :tada: status
- work around slack bug that prevented voter status block from updating


src/slack_interaction_handler.ts
```diff
@@ -267,6 +262,24 @@ export async function handleVoterStatusUpdate({
         );
       }
 
+      // HACK: work around slack bug updating blocks: we must first remove the dropdown before we change it
+      const oldBlock = payload.message.blocks[2];
```
tomerovadia: Is there a way to do this without hard-coding the 2? Maybe Andrew has a recommendation?

src/types.ts
```diff
@@ -12,8 +12,7 @@ export type VoterStatus =
   | 'IN_PERSON'
```
tomerovadia, line 12: What will happen if someone changes the dropdown of an old volunteer thread to ALREADY_VOTED (which will still be available in those dropdowns)?

src/slack_interaction_handler.ts
```diff
@@ -267,6 +262,24 @@ export async function handleVoterStatusUpdate({
...
+      payload.message.blocks[2] = {
+        type: 'section',
+        block_id: 'asdf',
+        text: {
+          type: 'mrkdwn',
+          text: 'Updating...',
+          verbatim: false,
+        },
+      };
+      await SlackInteractionApiUtil.replaceSlackMessageBlocks({
```
tomerovadia: I'm not sure I follow what's going on here. I know there's a bug but why are we calling replaceSlackMessageBlocks twice?

liewegas: Updating the selected option doesn't work. So we first change the block to "Updating..." and then change it back to the drop down, which is enough to coax slack into selecting the new item.

src/types.ts
liewegas: ah, good catch! fixed this, but i had to rebase this branch on top of a merge of the others to fix conflicts.. so i'll rebase/repush once the others (follow-up, unclaim) merge first.

src/slack_interaction_handler.ts
tomerovadia: I see. Have you heard back from Andrew regarding this?

src/slack_interaction_handler.ts
liewegas: I don't think so.. we want to update the block in the third slot. That's hard-coded throughout.  We could use the replaceVoterPanelBlocks helper to make the change, but the oldBlocks = line is still needed.

This is a workaround for a bug that will hopefully get fixed someday (@fongandrew?) so I'm not too concerned about it.  It is moved into a helper in a later PR.

fongandrew: I'm not sure I follow. What's the bug? The update isn't sticking?

fongandrew: Oh, I see what it is. Yeah, it's on our radar and we'll fix this within the next week or two hopefully. In the meantime, I _think_ changing the `block_id` on the new block to something random will unstick it. Assuming that's what it is.

liewegas: Yep, that worked! Thanks!

tomerovadia: Thanks @fongandrew! @liewegas can you create an issue on this repo to remind us to fix this at some point in the future, and link to the issue from the comment explaining why this ID is necessary?

tomerovadia: Regarding the hard-coding, I'm not seeing any other hardcoding like this in slack_interaction_handler -- are you seeing it throughout elsewhere? The only other place I'm seeing it is in slack_voter_status_message.

If you can create an issue in this repo to remind us to follow up on this another time I'm fine with submitting this as is, but hopefully we don't add too much more hard-coding.

tomerovadia (approved):

PR #207, opened 2020-11-12
add twilio_phone_number column to threads table

liewegas:
This is mostly for symmetry to make this table match all of the others.

We also take this opportunity to remove the catches that ignore errors when the threads table doesn't exist--it does, now, so we don't need them!


liewegas:
this is included in the sessions PR

PR #208, opened 2020-11-14
Support for helpline sessions

liewegas:
The key idea here is to separate different conversations with the helpline into different "sessions".  Sessions would most commonly be separated by elections, but they could also (in the future) be triggered if a long time has simply passed since the voter last interacted with the helpline.

When this code is deployed, all existing sessions will lack the sessionStartEpoch field, and will be considered "stale".  A stale session is a session that is in a sort of limbo state: we have decided that the sessions is over, but all of the blocks in slack don't reflect that fact.  This is because we'll have thousands and thousands of conversations leading up to an election, and will then do a mass-update in the database to, e.g., reset the voter status to Registered from Voted, but we don't currently iterate over all of the slack state to make it reflect that.

But, fear not--stale sessions show themselves as soon as you interact with them.

If a volunteer tries to text a voter via a stale session thread, they'll get an error message, and a helpful prompt to use one of the new commands (!resume-session or !new-session).  If a voter texts the helpline, they'll get the generic welcome text (as if they are a new user), and at that point the slack state will get refreshed so that the blocks says "This session has ended".

If a voter reenters the helpline, past sessions are visible in the history.  The existing history block is still there showing *just* messages for the new session (and whatever preceded HELPLINE).  Above that there are one or more messages that indicate there was a previous session and how long ago, along with buttons to Show or Hide it.  If the history is too long for a single slack message block, there are buttons to paginate.

If a volunteer encounters a stale thread, they have two choices:

- ``!resume-session`` will resume the existing session.  The blocks will refresh (pulling the latest voter status from the db), and the volunteer assignment will be unchanged.  The volunteer can then text the voter and the voter can respond as if nothing happened.
- ``!new-session channelname`` will open up a new session with the voter, create a new thread in an arbitrary channel (could be the same one!).  The voter state is refreshed and the volunteer assignment is cleared.  The old sessions history is of course visible.  The volunteer can immediately start texting the voter and the voter can respond, and the voter won't realize anything happened (no need to send HELPLINE to reenter), so the conversation can continue.

These two options, from the voter's perspective, look the same.  The only difference is on the volunteer side of things.  Is this a continuation of the earlier conversation, or a fresh conversation about a new upcoming election?  Is it more convenient to (1) clear the volunteer and (2) route the voter to a new channel all in one go?

There are a few other things that were added along the way that are unrelated to sessions:

- ``!route channelname`` now works in any voter thread.  This is way more convenient than the old ``@helpline ROUTER_VOTER cut-and-paste-id 48298 channelname`` flow.  The behavior is the same, including the check mark on success and error message in the thread if there is a problem.
- ``!fake-old-session`` will make a session *look* like a pre-sessions thread (with no sessionStartEpoch) for the purposes of testing.  This just cleared the field from redis.
- A series that adds ``twilio_phone_number`` to the ``threads`` table precedes all of this, which gets us away from relying on ``is_demo`` while the reset of the code uses user_id + twilio_phone_number.
- The voter header is shorter and generated from a single helper function.  (Push vs pull is uninteresting because it never changes within an org, and the userId + twilio number can share the same line.)

There are some other cosmetic improvements along the way (more X and white_check_mark reactions; links to new location of session/thread).

There is also some ick:

- Most of the existing postgres fields are TIMESTAMP and not TIMESTAMPTZ, which makes things gross with respect to timezone.  In order for conversion from postgres timestamp back to unix epoch to not offset by timezone offset each time, the new fields are TIMESTAMPTZ.  Probably all of the existing fields should have been, but they're not. In one case a select casts to ::timestamptz to work around this, but perhaps a future cleanup will convert everything to TIMESTAMPTZ... and also switch to UTC everywhere! But I leave that to future generations.
- Slack has a maximum message size of 3k-4k (not sure exactly), which prevents us from putting too much history content in one message. But our past session history is a single block that we update, so we have pagination.  We are also limited to like 5 buttons across, so it always shows first/last, current, and the next/prev. Even though the button content has a `type`, it only supports `plain_text`, not `mrkdwn`, so the current position looks like ``*2*`` (vs ``1`` and ``3`` for adjacent).  Clicking the current *2* button doesn't do anything, but you can't mix buttons and non-buttons in the blocks.  I don't think it matters.
- Also, if you expand blocks with lots of text, Slack does it's own "Show more" and "Show less" thing which is kind of annoying when you're literally clicking a "Show" button.. but I don't think there is a way around that.  It's only slightly annoying.

TODO:
- [x] fix router tests
- [x] note whether voter is new or returning (add returningVoter boolean to UserInfo)

A new thread with past sessions:

image

Expanded short session:

image

Expanded with pagination:

image

Volunteer texting to a stale session:

image


src/db_api_util.ts
```diff
@@ -581,25 +606,71 @@ export async function archiveDemoVoter(
...
+export async function setSessionEnd(
+  userId: string,
+  twilioPhoneNumber: string
+): Promise<void> {
+  logger.info(`ENTERING DBAPIUTIL.setSessionEnd`);
+  const client = await pool.connect();
+  try {
+    await client.query(
+      `UPDATE threads SET session_end_at = (
+        SELECT MAX(updated_at) + interval '1 second' FROM threads
```
benweissmann, line 618: What's the 1 second for?

src/router.ts
```diff
@@ -31,20 +31,57 @@ type UserOptions = {
...
   commandParentMessageTs: string;
+  commandMessageTs: string; // if command is in a thread, this is the in-thread message
+  commandChannel: string;
   routingSlackUserName: string;
   previousSlackChannelName: string;
 };
 
+export function isStaleSession(userInfo: UserInfo): boolean {
+  if (!userInfo.sessionStartEpoch) {
+    // legacy session from before 2020-11-03 election
```
benweissmann, line 42: This isn't specific to 2020-11-03, right? It's just any session that's been ended by running that SQL migration?

src/router.ts
```diff
@@ -1372,7 +1541,237 @@ export async function handleSlackVoterThreadMessage(
     `${userId}:${twilioPhoneNumber}`
   )) as UserInfo;
   // Only relay Slack messages from the active Slack thread.
-  if (userInfo.activeChannelId === reqBody.event.channel) {
+  if (
```
benweissmann, line 1844: Can we pull this stuff out into a separate function or file? This function is getting pretty hard to follow -- maybe something like `if (handleCommand(messageToSend)) { return; }` and then `handleCommand()` can return `true` if the message was a command and `false` if it was a normal message.

benweissmann: And then that function can help DRY out some of this, like handling the ACL check at the top of each of these branches

benweissmann (changes requested):
Looks good, just a few suggestions / questions

src/db_api_util.ts
liewegas: The message might be 11.2 seconds, the updated_at 11.3 seconds, and when we do the select it ends up rounded down to 11, which means we leave off the the last message in the session.  The 1 second just provides a buffer.

src/router.ts
liewegas: It's specific to anything that predates this branch being merged, since starting now we'll populate the sessionStartEpoch when creating a new UserInfo.

src/router.ts
liewegas: Yep--updated!

src/db_api_util.ts
benweissmann: got it -- can you just add a comment to that effect?

benweissmann (approved):
looks great!

PR #210, opened 2020-11-15
Handle VOTED in more places

liewegas:
- If the user texts VOTED, we always update the voter status
- We continue to autoreply before HELPLINE (as before)
- New: autoreply if we are in the lobby before the voter has intervened
- New: autoreply if we would have sent a "Welcome back! Finding a volunteer..." message

This is based on top of #205 
This was already tested in staging

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -6,7 +6,6 @@ import morgan from 'morgan';
 import axios, { AxiosResponse } from 'axios';
 import { Pool } from 'pg';
 
-import * as MessageConstants from './message_constants';
 import * as SlackApiUtil from './slack_api_util';
 import * as SlackInteractionHandler from './slack_interaction_handler';
 import * as TwilioApiUtil from './twilio_api_util';
@@ -30,8 +29,6 @@ import { handleTwilioStatusCallback } from './twilio_status_callback_handler';
 import { SlackInteractionEventPayload } from './slack_interaction_handler';
 import { TwilioRequestBody } from './twilio_util';
 import * as KeywordParser from './keyword_parser';
-import { SlackActionId } from './slack_interaction_ids';
-import { VoterStatus } from './types';
 
 const app = express();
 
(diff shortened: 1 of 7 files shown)


src/app.ts
```diff
@@ -356,6 +359,58 @@ const handleIncomingTwilioMessage = async (
       logger.info(
         `SERVER.handleIncomingTwilioMessage (${userId}): Voter has previously confirmed the disclaimer, or one is not required for this organization.`
       );
+
```
tomerovadia, line 359: I intended `handleIncomingTwilioMessage` to be high-level logic for which response strategy a voter should receive (e.g. handleClearedVoter, clarifyHelplineRequest), and for app.ts to handle only high-level logic or logic that's commonly needed across each of these strategies (e.g. `populateIncomingDbMessageTwilioEntry`).

To keep this design pattern and also prevent `handleIncomingTwilioMessage` from getting too much bigger, can you find a better place to put this logic?

src/app.ts
```diff
@@ -356,6 +359,58 @@ const handleIncomingTwilioMessage = async (
...
+        });
+        await SlackApiUtil.sendMessage(
+          `*Operator:* Voter status changed to *VOTED* by user text.`,
+          {
+            channel: userInfo.activeChannelId,
+            parentMessageTs: userInfo[userInfo.activeChannelId],
+          }
+        );
+
+        // update blocks
```
tomerovadia: capitalize comments

src/slack_interaction_handler.ts
```diff
@@ -265,12 +260,12 @@ export async function handleVoterStatusUpdate({
         );
       }
 
-      // Replace the entire block so that the initial option change persists.
-      await SlackInteractionApiUtil.replaceSlackMessageBlocks({
-        slackChannelId: payload.channel.id,
-        slackParentMessageTs: payload.container.thread_ts,
-        newBlocks: payload.message.blocks,
-      });
+      // Update the voter block
```
tomerovadia: What's the thinking behind this comment change?

src/slack_interaction_handler.ts
```diff
@@ -112,12 +112,7 @@ const getClosedVoterPanelText = (
   // See https://api.slack.com/reference/surfaces/formatting#visual-styles
   const specialSlackTimestamp = `<!date^${timeSinceEpochSecs}^{date_num} {time_secs}|${new Date()}>`;
 
-  switch (selectedVoterStatus) {
-    case 'VOTED':
-      return `*Congratulations!* :tada: This voter was marked as *VOTED* by *${originatingSlackUserName}* on *${specialSlackTimestamp}*. On to the next one! :ballot_box_with_ballot:`;
-    default:
-      return `:no_entry_sign: This voter was marked as *${selectedVoterStatus}* by *${originatingSlackUserName}* on *${specialSlackTimestamp}*. :no_entry_sign:`;
```
tomerovadia: Are you removing the spam/refused notification?

src/app.ts
liewegas: Okay, I pushed a commit that moves this block and the large one below into helpers in router.ts.  The calls still need to stay here in app.ts because of the high-level behavior: (1) always log, and (2) reply to voted if we aren't cleared yet.

src/slack_interaction_handler.ts
liewegas: None really--the original comment just got moved into the helper.  Switched it back in the latest commit.

src/slack_interaction_handler.ts
liewegas: That hunk was in the other PR this was based on top of.  The change was removing the switch statement because there is no longer a 'voted' blocks that hides all the buttons; refused/spam still behave as before.

src/app.ts
```diff
@@ -378,6 +385,15 @@ const handleIncomingTwilioMessage = async (
           inboundDbMessageEntry,
           twilioCallbackURL
         );
+        // Voter texted VOTED during the state determination
```
tomerovadia, line 391: Any logic pertaining to VOTED needs to be flag-guarded to only VOTE_AMERICA

src/slack_interaction_handler.ts
```diff
@@ -260,18 +260,12 @@ export async function handleVoterStatusUpdate({
         );
       }
 
-      // HACK: work around slack bug updating blocks by adjusting the block_id
-      payload.message.blocks[2]['block_id'] = Math.random()
-        .toString(36)
-        .replace(/[^a-z]+/g, '')
-        .substr(0, 8);
-
       // Replace the entire block so that the initial option change persists.
```
tomerovadia: You change `replaceSlackMessageBlocks` with `updateVoterStatusBlocks` but the comment above still says "Replace the entire block..."

tomerovadia (commented):
Feel free to merge

tomerovadia (approved):

PR #212, opened 2020-11-16
Normalize creation of "voter header"

liewegas:
(the text identifying the voter in the channel message, preceding the volunteer and voter status blocks).

- Put it all in a helper
- 'New' vs 'Returning' voter
- optional suffix (e.g., 'known phone number', or 'routed from __ by __')


liewegas:
Included this in sessions pr

PR #214, opened 2020-11-16
Session topics

liewegas:
The idea here is to record which questions were asked or topics discussed during a helpline session.  This can be used for later data analysis.

We add a new table that is keyed off of the *session* (user_id, twilio_phone_number, session_start_at).  It includes a jsonb column that has a JSON array of topics.

The voter header in slack now contains a multi-select that lets you pick one or more (or no) topics.  This block is hidden if the voter is marked Refused or Spam, and restored on Undo.  The slack multi-selector isn't the greatest UX (you can't see what is selected unless you open it up; another click is needed before you see available options; two more clicks after making your pick to close the dialog), but it's what we have available.

If there are past session for the voter, the topics dicussed are include in the history message.

image
image
image

Edge cases / QA tests:

- The # of selected items should update in both the thread and the channel
- If you do refused or spam and then undo, when the block reappears it should show the same selected items
- Routing a voter between channels should preserve the topics.  Same goes for a voter in the lobby who has topics selected and then texts their state and gets routed automatically

### DIFF
--- a/sql/20201116-voter-topics.sql
+++ b/sql/20201116-voter-topics.sql
@@ -1,0 +1,13 @@
+--- record voter topic selections
+
+CREATE TABLE session_topics (
+  created_at TIMESTAMPTZ NOT NULL,
+  user_id TEXT NOT NULL,
+  user_phone_number TEXT NOT NULL,
+  twilio_phone_number TEXT NOT NULL,
+  session_start_at TIMESTAMPTZ NOT NULL,
+  is_demo BOOLEAN,
+  archived BOOLEAN,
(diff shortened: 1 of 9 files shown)


liewegas:
@tomerovadia I spent some time testing this on staging after merging and found one issue with how Undo interacts with one of the other PRs, but it was simplest to fix it in this PR.  Bonus is that !resume-session no longer duplicates the code that restores/rebuilds the voter panel.  I think this one is ready to merge.

tomerovadia (approved):

PR #215, opened 2020-11-17
Create Route to Journey message shortcut

tomerovadia:
Based on top of #214 Session topics

Technical notes:
- Since VoteAmerica is not using the PUSH functionality, Push will be used to indicate Journey. This will generate new sets of open pods that can be used to track and easily update which pods can receive additional voters for journey.
- I needed to change the LoadBalancer to accept regions as state names. Usually the load balancer only handles completely new voters, which are routed based on their state name. But because a voter might have been routed to a different state before being routed to journey, we need to parse their state name from the Slack channel name, and accommodate situations in which this is not a state but a region name.
- Push/Pull has a completely different purpose for other organizations, but because VoteAmerica doesn't use the push open channel lists at all, the lists can be conveniently repurposed for the concept of Journey instead. Note: All push paths in the router are still inactive, as there are no phone numbers associated with push in the original sense.
- The routeVoterToSlackFunction usually takes an adminCommandParams argument that includes information about the message of the admin command. This PR accepts a third way to route a voter, which includes the adminCommandParams but without the message details (since the routing mechanism is a message shortcut instead of a command).

This feature has two objectives:
- Leverage volunteers who are especially highly trained and knowledgeable about voting rules to focus on incoming voter questions, which are likely the most difficult questions and most important to helping the voter successfully vote. This role is **frontline** volunteer.
- Systematize follow-up with voters by having volunteers "adopt" and "shepherd" a group of voters (e.g. 20 each) until they successfully vote. This role is **journey** volunteer.

Each pod will be designated as a "frontline" or "journey" pod, with all volunteers in a pod having one of the two above responsibilities. Journey pods will be closed to the traditional entrypoint and thus not receive any new voters. Instead, frontline pods will receive new voters, and when a volunteer is done with a conversation, they should mention to the voter that they are connecting them with a volunteer who will be available to answer any additional questions that come up through Election Day.

This feature will be a message shortcut Route to Journey. In the normal use case, a volunteer will be able to select this shortcut on any message within a voter's thread. A modal will appear to confirm this is the action the volunteer wants to take. If confirmed, the subsequent operations will:
1. Determine which journey pods (Slack channels) are open for this state as well as the state of the counter that round robins voters to these pods. This helps determine which pod the voter should be routed to.
2. The remaining operations should be the equivalent of a manual admin action to route the voter to the selected pod.

Edge cases / QA tests:
- If a volunteer selects the shortcut on any message outside of a voter thread, a modal appears instructing them on proper usage.
- If a volunteer selects Route to Journey for a voter thread that is already listed as a journey pod for that same U.S. state, if the same pod is selected, an error modal appears rather than routing the voter to the same channel.
- If no journey pod is open, the volunteer receives a modal letting them know.
- If a voter is in a region channel, they should be routed to a journey pod that corresponds to that region.
- If a voter selects U.S. state A and then is routed to U.S. state B and is then routed to journey, they should be routed to a state B journey pod (not state A).
- Trying to Route to Journey a voter that is in a national pod should return a modal that says there are no open journey pods for this state.
- Make sure I didn't break any of the other message shortcuts (in particular Reset Demo).

### DIFF
--- a/src/async_jobs.ts
+++ b/src/async_jobs.ts
@@ -502,10 +502,13 @@ async function slackInteractivityHandler(
       }
 
       case SlackCallbackId.ROUTE_TO_JOURNEY: {
-        const modalPrivateMetadata = JSON.parse(
+        const modalPrivateMetadata = SlackInteractionHandler.parseSlackModalPrivateMetadata(
           payload.view.private_metadata
-        ) as SlackModalPrivateMetadata;
-        if (modalPrivateMetadata.commandType !== 'ROUTE_TO_JOURNEY') {
+        );
+        if (
+          modalPrivateMetadata &&
+          modalPrivateMetadata.commandType !== 'ROUTE_TO_JOURNEY'
+        ) {
           throw new Error(
             `Got callback ID ROUTE_TO_JOURNEY but private commandType was ${modalPrivateMetadata.commandType}`
           );
--- a/src/async_jobs.ts
+++ b/src/async_jobs.ts
@@ -243,12 +243,58 @@ async function slackInteractivityHandler(
           );
           await SlackApiUtil.updateModal(viewId, slackView);
           logger.info(
-            `SLACKINTERACTIONHANDLER.receiveResetDemo: Volunteer used a shortcut on an invalid message.`
+            `ASYNCJOBS.slackInteractivityHandler: Volunteer used a reset demo shortcut on an invalid message.`
           );
           return;
         }
 
         await SlackInteractionHandler.receiveResetDemo({
+          payload,
+          redisClient,
+          modalPrivateMetadata,
+          twilioPhoneNumber: redisData.twilioPhoneNumber,
+          userId: MD5.hex(redisData.userPhoneNumber),
+          viewId,
+        });
+        return;
+      }
+
+      case SlackCallbackId.ROUTE_TO_JOURNEY: {
+        const { viewId } = interactivityMetadata;
+        if (!viewId) {
+          throw new Error(
+            'slackInteractivityHandler called for message_action without viewId'
+          );
+        }
+
+        const MD5 = new Hashes.MD5();
+
+        // Ignore Prettier formatting because this object needs to adhere to JSON strigify requirements.
+        // prettier-ignore
+        const modalPrivateMetadata = {
+          "commandType": 'ROUTE_TO_JOURNEY',
+          "userId": redisData ? MD5.hex(redisData.userPhoneNumber) : null,
+          "twilioPhoneNumber": redisData ? redisData.twilioPhoneNumber : null,
+          "originatingSlackUserName": originatingSlackUserName,
+          // destinationSlackChannelName is populated later
+        } as SlackModalPrivateMetadata;
+
+        if (!originatingSlackChannelName || !redisData) {
+          modalPrivateMetadata.success = false;
+          modalPrivateMetadata.failureReason = 'invalid_shortcut_use';
+          await DbApiUtil.logCommandToDb(modalPrivateMetadata);
+          const slackView = SlackBlockUtil.getErrorSlackView(
+            'not_active_voter_parent_thread',
+            'This shortcut is not valid on this message.'
+          );
+          await SlackApiUtil.updateModal(viewId, slackView);
+          logger.info(
+            `ASYNCJOBS.slackInteractivityHandler: Volunteer used a route to journey shortcut on an invalid message.`
+          );
+          return;
+        }
+
+        await SlackInteractionHandler.receiveRouteToJourney({
           payload,
           redisClient,
           modalPrivateMetadata,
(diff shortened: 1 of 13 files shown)


src/async_jobs.ts
```diff
@@ -243,6 +259,52 @@ async function slackInteractivityHandler(
...
+        const { viewId } = interactivityMetadata;
+        if (!viewId) {
+          throw new Error(
+            'slackInteractivityHandler called for message_action without viewId'
+          );
+        }
+
+        const MD5 = new Hashes.MD5();
+
+        // Ignore Prettier formatting because this object needs to adhere to JSON strigify requirements.
```
benweissmann, line 272: I'm not sure what this means -- Prettier doesn't change the semantics of your code; there shouldn't be any runtime difference between prettier code and non-prettier code.

src/slack_interaction_handler.ts
```diff
@@ -997,12 +1174,129 @@ export async function receiveResetDemo({
...
+    if (!userInfo) {
+      modalPrivateMetadata.success = false;
+      modalPrivateMetadata.failureReason = 'no_user_info';
+      await DbApiUtil.logCommandToDb(modalPrivateMetadata);
+      throw new Error(
+        `SLACKINTERACTIONHANDLER.receiveRouteToJourney: Interaction received for voter who has redisData but not userInfo: active redisData key is ${payload.channel.id}:${payload.message.ts}, userInfo key is ${redisUserInfoKey}.`
       );
     }
 
+    if (!(payload.channel.id === userInfo.activeChannelId)) {
```
benweissmann, line 1259: nit: probably easier to read as `payload.channel.id !== userInfo.activeChannelId`

src/async_jobs.ts
```diff
@@ -414,6 +501,22 @@ async function slackInteractivityHandler(
         return;
       }
 
+      case SlackCallbackId.ROUTE_TO_JOURNEY: {
+        const modalPrivateMetadata = JSON.parse(
```
benweissmann: I'd recommend explicitly checking the type of this data rather than assuming that it will be valid JSON -- maybe something like `parseSlackModalPrivateMetadata(metadata: string): SlackModalPrivateMetadata`. So that we get better errors if we accidentally open a modal with invalid data, or if the expected data changes in between versions of the helpline and someone opens a modal, then we deploy new code, and then they click a button in the modal.

benweissmann (changes requested):
This looks good! Just a couple suggestions.

src/slack_interaction_handler.ts
tomerovadia: thanks!

src/async_jobs.ts
tomerovadia: This is copied from RESET_DEMO and it's been a while but IIRC the double quotes are necessary for the keys (prettier removes them) in order for JSON.stringify to work, for some reason. I can look up the documentation I found previously again if necessary.

src/async_jobs.ts
tomerovadia: Done. I hope this is what you had in mind.

src/async_jobs.ts
benweissmann: I would give this another try- the quotes around the key names are just used during parsing when the code is loaded; they're not going to affect how the code works.

benweissmann (approved):
Looks good. I would give it another try without ignoring prettier; I am almost certain that quoting the fields does not make any difference at run-time.

src/slack_interaction_handler.ts
```diff
@@ -1193,6 +1196,124 @@ export async function receiveResetDemo({
...
+            'route_to_journey_error_no_pods_found',
+            `No journey pods found for U.S. state name: *${activeStateName}*. Please contact an admin.`
+          );
+        } else if (
+          activeSlackChannelName === selectedSlackChannelName
+        ) {
+          slackView = SlackBlockUtil.getErrorSlackView(
+            'route_to_journey_error_routing_to_same_channel',
+            `The voter is already in an open journey pod for this U.S. state.`
+          );
```
liewegas, line 1303: The condition and error message don't completely match. IIUC this basically means that if you have multiple possible target channels for the state, and the voter is already in one of them, then sometimes we'll randomly pick the same one and get this error message (which will make sense in that context) or we'll randomly pick a different journey pod and the voter will get routed there.  If there is a way to identify whether the current channel is a journey channel that would be better.  If not, this should probably go on a list of things to improve later!

liewegas (approved):
This looks great! Just one flag on the routing errors.

src/slack_interaction_handler.ts
tomerovadia: Ya you're describing that accurately. This is something worth following up on -- unfortunately I didn't see an easy way to do this without adding complexity to the LoadBalancer or a second Redis fetch for available pods. I think for the Georgia election we can just keep this in mind (this hopefully shouldn't happen often) -- and at least the case of routing to the same channel will be caught and stopped.

I created an issue so we can follow-up: https://github.com/tomerovadia/VoterHelpLine/issues/217

If everything else looks good do you mind merging this once #214 Session topics is in?

PR #224, opened 2020-12-03
Change needs-attention handling

liewegas:
- Needs attention no longer depends on whether a volunteer is assigned.
- Needs attention is set on a new thread, and when an incoming message is handled from a cleared voter (i.e., past lobby, or volunteer engaged)
- Needs attention is cleared when a slack thread message is relayed to the voter (i.e., not a command)

Implementation notes:

- We pull the updates out of the twilio/slack send functions, which are deep in the call chain and kludgey. Instead, we update from high level methods.
- The thread updated_at field is now updated from logMessageToDb.
- The old method that updates from the databaseMessageEntry is gone; instead the callers explicitly set true or false for the thread.

This is simpler and easier to understand. The main net change from a volunteer perspective is that if you start talking to a voter that is unclaimed their needs-attention status will get cleared. However, they will still be unclaimed, so I don't think they will get lost.  We *could* add back in the unclaimed-always-needs-attention behavior in here if we want to, though, but I think it's slightly less intuitive...


liewegas:
Brenna approves of behavior change.

benweissmann (approved):

PR #238, opened 2020-12-16
Allow keywords that route directly to states

liewegas:
This lets you set keywords in redis to route directly to a state.  it
does not allow for any customized messages to go with that, however.

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -443,6 +443,27 @@ async function introduceNewVoterToSlackChannel(
   return response.data.ts;
 }
 
+export async function parseStateKeyword(
+  redisClient: PromisifiedRedisClient,
+  userInfo: UserInfo,
+  userMessage: string
+): Promise<boolean> {
+  // Is this a special keyword?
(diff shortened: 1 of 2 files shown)


src/router.ts
```diff
@@ -443,6 +443,27 @@ async function introduceNewVoterToSlackChannel(
   return response.data.ts;
 }
 
+export async function parseStateKeyword(
+  redisClient: PromisifiedRedisClient,
+  userInfo: UserInfo,
+  userMessage: string
+): Promise<boolean> {
+  // Is this a special keyword?
+  const keywords = await RedisApiUtil.getHash(redisClient, 'keywordConfig');
```
benweissmann: Do we want this in redis? I think this is unlikely to change frequently, so it feels like an env var or a configuration in the code based on CLIENT_ORGANIZATION would be easier to manage.

liewegas: Yeah, that's simpler.  Updated

benweissmann (approved):

PR #242, opened 2020-12-26
Tweaks to journey/frontline copy; create new thread for GA Dems returning voters

benweissmann:
- Consistently use the journey/frontline terminology rather than push/pull when changing entrypoints
- GA Dems wants a new thread when a voter returns after >24 hours. This modifies the "welcome back" behavior for GA dems to create a new thread by "routing" back to the same channel.

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -802,16 +802,25 @@ export async function routeVoterToSlackChannel(
     );
     // Operations for AUTOMATED route of voter.
   } else if (!skipLobby && userInfo.activeChannelId != 'NONEXISTENT_LOBBY') {
-    await SlackApiUtil.sendMessage(
-      `*Operator:* Routing voter to ${SlackApiUtil.linkToSlackChannel(
+    let msg;
+    if (userInfo.activeChannelId === destinationSlackChannelId) {
+      // we're routing back to the same channel (to make a new thread)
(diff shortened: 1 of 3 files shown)


src/router.ts
```diff
@@ -1415,6 +1424,16 @@ export async function handleClearedVoter(
...
+
+    if (process.env.CLIENT_ORGANIZATION === 'GADEMS') {
+      // GA Dems starts a new thread if an old voters comes back rather
+      // than continuing to use the existing thread. We do this by "routing"
+      // the voter back to the same channel
+      await routeVoterToSlackChannel(userInfo, redisClient, {
+        twilioPhoneNumber,
+        destinationSlackChannelName: userInfo.activeChannelName,
+      });
+    }
```
liewegas, line 1436: This will make the voter pop up again in the channel, but it won't be a new "session" in that the voter history from the previous day(s) wont' be collapsed. That might be fine for their use-case though?

liewegas (approved):

src/router.ts
benweissmann: yeah, i think this is desirable for ga dems (we're only looking at ~2 weeks total of interactions so i think it's fine for it all to be inline)

PR #244, opened 2020-12-27
More GADEMS customizations

benweissmann:
Similar to VOTE_AMERICA, GADEMS wants a more casual intro without a formal disclaimer. So i've updated GADEMS to behave like VOTE_AMERICA but always routing to Georgia.

### DIFF
--- a/src/message_constants.ts
+++ b/src/message_constants.ts
@@ -46,7 +46,7 @@ export function WELCOME_VOTER(): string {
     case 'VOTER_HELP_LINE':
       return 'Welcome to Voter Help Line! We are excited to help you vote.\n\nPlease note that this is not an official or government-affiliated service. Volunteers will do their best to share official links that support their answers to your questions, but by using this service you release Voter Help Line of all liability for your personal voting experience.\n\nReply AGREE to confirm that you understand and would like to continue. (Msg & data rates may apply).';
     case 'GADEMS':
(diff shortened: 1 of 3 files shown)


src/message_constants.ts
```diff
@@ -46,7 +46,7 @@ export function WELCOME_VOTER(): string {
     case 'VOTER_HELP_LINE':
       return 'Welcome to Voter Help Line! We are excited to help you vote.\n\nPlease note that this is not an official or government-affiliated service. Volunteers will do their best to share official links that support their answers to your questions, but by using this service you release Voter Help Line of all liability for your personal voting experience.\n\nReply AGREE to confirm that you understand and would like to continue. (Msg & data rates may apply).';
     case 'GADEMS':
-      return 'Welcome to Voter Help Line! We are excited to help you vote.\n\nPlease note that this is not an official or government-affiliated service. Volunteers will do their best to share official links that support their answers to your questions, but by using this service you release us of all liability for your personal voting experience.\n\nReply AGREE to confirm that you understand and would like to continue. (Msg & data rates may apply).';
+      return 'Hi! You’ve reached the Georgia Voter Assistance Textline.\n\nReply HELPLINE to be connected to a volunteer.';
```
liewegas: It's a bit odd that we call it the 'textline' but in the next sentence ask them to reply helpline.

liewegas (approved):
I think 'textline' should read 'helpline'.. otherwise looks good!

